PR #1, opened 2018-07-26
Check on usage of partial functions

vrom911:
On current test file I got this output:

```
!!! Found 1 warning(s) in test/Spec.hs
Warning: Usage of the partial function ''head'
main = print $ head ["Hello"]
               ^^^^
```

### DIFF
--- a/src/Stan.hs
+++ b/src/Stan.hs
@@ -1,7 +1,35 @@
+-- | This module introduces GHC plugin for static analysis @Stan@.
+
 module Stan
-       ( someFunc
+       ( plugin
        ) where
 
-someFunc :: IO ()
-someFunc = putStrLn ("someFunc" :: String)
+import Control.Monad (unless)
+import Control.Monad.IO.Class (liftIO)
+import Data.Foldable (for_)
 
+import FastString (unpackFS)
+import HsDecls (HsGroup)
+import HsExtension (GhcRn)
+import Plugins (CommandLineOption, Plugin (..), defaultPlugin)
+import SrcLoc (srcSpanFile)
+import TcRnTypes (TcGblEnv (..), TcM)
+
+import Stan.Analysis (stan)
+import Stan.Warning (showWarning)
+
+plugin :: Plugin
+plugin = defaultPlugin { renamedResultAction = stanPlugin }
+
+stanPlugin :: [CommandLineOption] -> TcGblEnv -> HsGroup GhcRn -> TcM (TcGblEnv, HsGroup GhcRn)
+stanPlugin _ tcEnv hsGroup = do
+    -- let modulePath = ms_hspp_file modSummary
+    let warnings = stan hsGroup
+    unless (null warnings) $ do
+        -- print the module name
+        let fileName = unpackFS $ srcSpanFile $ tcg_top_loc tcEnv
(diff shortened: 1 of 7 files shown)


vrom911:
I've patched output a little bit. Now it looks like this:

```
test/Spec.hs: Found 1 warning(s)
--------------------

(4:16) Warning: Usage of the partial function 'head'

  |
4 | main = print $ head ["Hello"]
  |                ^^^^

```

src/Stan/Partial.hs
```diff
@@ -31,26 +31,33 @@ findPartials = mapMaybe toWarning . lNames
     toWarning (L l name) =
         let modName = moduleName $ nameModule name
             occName = nameOccName name
-        in
-        if checkName modName occName
-            then Just $ Partial (occNameString occName) l
-            else Nothing
+        in guard (ifPartial modName occName && isRealSrcSpan l)
+               *> Just (Partial (occNameString occName) $ getReal l)
```
ChShersh: Programming pattern like `if isJust foo then fromJust foo` is considered bad. Here you can use `do`-notation desugaring and the fact that pattern-match failure results in call to `fail` function. In case of `Maybe` function `fail` just returns `Nothing`. So you can safely write code like this:

```haskell
foo = do
    RealSrcSpan src <- Just l
    guard $ isPartial modName occName
    Just {- or pure -} $ Partial (occNameString occName) src
``` 

Alternatively you can change type of `getReal` function to be `getReal :: SrcSpan -> Maybe RealSrcSpan` but approach with `fail` allows you to not write extra function.

src/Stan/Warning.hs
```diff
@@ -3,29 +3,63 @@
...
+
+@
+src/Hello.hs: Found 3 warning(s)
+--------------------
+@
+-}
+moduleOverall :: [Warning] -> String -> String
+moduleOverall ws name = unlines
+    [ name ++ ": Found " ++ show (length ws) ++ " warning(s)"
+    , replicate 20 '-'
```
ChShersh: Probably you can move `name ++ ": Found " ++ show (length ws) ++ " warning(s)"` into variable like `header` so later you can just do `replicate (length header) '-'` instead of using hardcoded number `20`.

src/Stan.hs
```diff
@@ -25,11 +25,11 @@ stanPlugin :: [CommandLineOption] -> TcGblEnv -> HsGroup GhcRn -> TcM (TcGblEnv,
...
     -- let modulePath = ms_hspp_file modSummary
     let warnings = stan hsGroup
-    unless (null warnings) $ do
+    unless (null warnings) $ liftIO $ do
         -- print the module name
         let fileName = unpackFS $ srcSpanFile $ tcg_top_loc tcEnv
-        liftIO $ putStrLn $ "!!! Found " ++ show (length warnings) ++ " warning(s) in " ++ fileName
-        fileContent <- liftIO $ readFile fileName
-        for_ warnings (liftIO . putStrLn . showWarning fileContent)
+        putStrLn $ moduleOverall warnings fileName
```
ChShersh: It was a good idea to move this into separate function. Sometimes functions are needed not only to reduce code duplication but to improve code readability and to assign name (description) to some chunk of code 👍

PR #12, opened 2020-04-10
[#9] Count the number of analyzed modules

### LINKED ISSUE #9: Count the number of analyzed modules
Also, how should we display this information in the terminal? I guess, at the beginning of the output is fine.

chshersh:
Resolves #9

### DIFF
--- a/src/Stan/Analysis.hs
+++ b/src/Stan/Analysis.hs
@@ -12,6 +12,7 @@ module Stan.Analysis
     ) where
 
 import HieTypes (HieFile)
+import Relude.Extra.Lens (Lens', lens, over)
 
 
 {- | This data type stores all information collected during static analysis.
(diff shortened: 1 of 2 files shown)


src/Stan/Analysis.hs
```diff
@@ -0,0 +1,51 @@
...
+initialAnalysis :: Analysis
+initialAnalysis = Analysis
+    { analysisModuleCount    = 0
+    , analysisLinesOfCode    = 0
+    , analysisUsedExtensions = 0
+    , analysisObservations   = ()
+    }
+
+-- QUESTION: is it better to create a newtype? Also, help with the name...
+type AnalyseM a = State Analysis a
```
vrom911, line 35: For the name, I would prefer a more specific one, like `AnalysisState`. But this then raises the question, do we need the type alias/newtype then?

src/Stan/Analysis.hs
```diff
@@ -0,0 +1,51 @@
...
+
+incModuleCount :: AnalyseM ()
+incModuleCount = modify' $
+    \analysis -> analysis { analysisModuleCount = analysisModuleCount analysis + 1 }
+    -- we probably will use lenses for this kind of stuff in the future...
```
vrom911, line 40: Sounds like a nice idea! :+1:

vrom911: We can start with the provided by `relude` lens module, and see if it is enough for our needs (should be)

src/Stan/Analysis.hs
chshersh: @vrom911 Good point :+1: I don't mind expanding type alias explicitly

PR #15, opened 2020-04-10
[#4] Read hie files of the target project

### LINKED ISSUE #4: hie files for the target project
Should we assume that they are present? Or should we create them during the run?
In the first option, we should add instructions, but with the second option, we need to decide how to make it more efficiently.

vrom911:
Resolves #4

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,3 +5,29 @@
 [![MPL-2.0 license](https://img.shields.io/badge/license-MPL--2.0-blue.svg)](https://github.com/kowainik/stan/blob/master/LICENSE)
 
 Haskell static analyser
+
+## Usage
+
+Stan is working with the `.hie` files to analyse Haskell projects. Therefore,
+Stan requires to have generated `hie` files provided by the users. Fortunately,
+it is straightforward to satisfy this necessity. You can build the project with
+special GHC options:
+
+```shell
+$ cabal v2-build \
+  --ghc-options=-fwrite-ide-info \
+  --ghc-options=-hiedir=.hie
+```
+
+Or you can also add these options to your project's `.cabal` file, so you will
+always have up to date `.hie` files:
+
+```haskell
+    ghc-options:       -fwrite-ide-info
+                       -hiedir=.hie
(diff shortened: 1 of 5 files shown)


src/Stan.hs
```diff
@@ -10,8 +10,13 @@ module Stan
...
+import Stan.Cli (CliArgs (..), runStanCli)
+import Stan.Hie (readHieFiles)
 
 
 runStan :: IO ()
-runStan = runStanCli >>= \_ -> pass
+runStan = runStanCli >>= \CliArgs{..} -> do
+    hieFiles <- readHieFiles cliArgsHiedir
+    let _analysis = runAnalysis hieFiles
+    pass
```
chshersh, line 22: You can do `print` analysis for now to check how it works :slightly_smiling_face:

README.md
```diff
@@ -5,3 +5,29 @@
...
+
+```haskell
+    ghc-options:       -fwrite-ide-info
+                       -hiedir=.hie
+```
+
+> _Note:_ here we recommend to generate the `hie` files into `.hie/` folder. As
+> it is the recommendation only, you can specify your own folder as well. But
+> then you will need to run `stan` using the `--hiedir` option with the
+> specified path to your `hie` folder.
```
chshersh, line 33: Juicy documentation :kiwi_fruit:

src/Stan/Hie.hs
```diff
@@ -0,0 +1,36 @@
...
+
+
+{- | Returns contents of all @.hie@ files recursively in the given
+@hie@ directory.
+-}
+readHieFiles :: FilePath -> IO [HieFile]
+readHieFiles hieDir = do
+    nameCache <- createNameCache
+    hieContent <- getDirRecursive hieDir
+    hieFiles <- filterM doesFileExist hieContent
```
chshersh, line 28: Hmm... I think if we also filter by the `.hie` extension, this code should easily work with the `dist-newstyle` and `.stack-work` directories. What do you think?

PR #18, opened 2020-04-12
[#17] Create printing debugging functions for HieFile

### LINKED ISSUE #17: Create printing debugging functions for HieFile

chshersh:
Resolves #17

### DIFF
--- a/src/Stan/Hie/Debug.hs
+++ b/src/Stan/Hie/Debug.hs
@@ -1,0 +1,79 @@
+{-# LANGUAGE StandaloneDeriving #-}
+{-# OPTIONS_GHC -fno-warn-orphans #-}
+
+{- |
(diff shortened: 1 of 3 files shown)


src/Stan/Hie/Debug.hs
```diff
@@ -0,0 +1,79 @@
...
+deriving stock instance Show a => Show (FieldLbl a)
+
+instance Show Module where
+    show = moduleStableString
+
+instance Show ModuleName where
+    show = moduleNameString
+
+instance Show Name where
+    show = nameStableString
```
vrom911, line 79: Heh, I wonder how come it is not in GHC :thinking:

chshersh: I also wonder... It seems so easy to implement and so useful to have at the same time :thinking:

PR #19, opened 2020-04-12
[#16] Create core data types

### LINKED ISSUE #16: Create core data types
`Inspection` and `Observation`.
See [Glossary](https://github.com/kowainik/stan/wiki/Glossary) for more info about the meaning.

vrom911:
Resolves #16

### DIFF
--- a/src/Stan/Hie/Debug.hs
+++ b/src/Stan/Hie/Debug.hs
@@ -1,5 +1,6 @@
-{-# LANGUAGE StandaloneDeriving #-}
 {-# OPTIONS_GHC -fno-warn-orphans #-}
+
+{-# LANGUAGE StandaloneDeriving #-}
 
 {- |
 Copyright: (c) 2020 Kowainik
--- a/src/Stan/Inspection.hs
+++ b/src/Stan/Inspection.hs
@@ -26,7 +26,7 @@ data Inspection = Inspection
     , inspectionName        :: !Text
     , inspectionDescription :: !Text
     , inspectionSolution    :: !Text
-    , inspectionCategory    :: ![Category]
+    , inspectionCategory    :: !(NonEmpty Category)
     , inspectionSeverity    :: !Severity
     } deriving stock (Show)
 
--- a/src/Stan/Inspection.hs
+++ b/src/Stan/Inspection.hs
@@ -1,0 +1,47 @@
+{- |
+Copyright: (c) 2020 Kowainik
+SPDX-License-Identifier: MPL-2.0
+Maintainer: Kowainik <xrom.xkov@gmail.com>
+
+__Inspection__ — check or test provided by Stan.
+-}
+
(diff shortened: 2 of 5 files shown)


src/Stan/Hie/Debug.hs
```diff
@@ -1,6 +1,7 @@
-{-# LANGUAGE StandaloneDeriving #-}
```
chshersh, line 1: Thanks for fixing the style :pray:

src/Stan/Inspection.hs
```diff
@@ -0,0 +1,47 @@
...
+
+-- | A type of the inspection.
+newtype Category = Category
+    { unCategory :: Text
+    } deriving newtype (Show)
+
+-- | Severity level of the inspection.
+data Severity
+    = Severe
+    | NotReallySevere
```
chshersh, line 41: :rofl:

src/Stan/Inspection.hs
```diff
@@ -0,0 +1,47 @@
...
+
+{- | Data type that represents a check/test, or how we call it
+__inspection__ that is provided by the Stan tool.
+-}
+data Inspection = Inspection
+    { inspectionId          :: !(Id Inspection)
+    , inspectionName        :: !Text
+    , inspectionDescription :: !Text
+    , inspectionSolution    :: !Text
+    , inspectionCategory    :: ![Category]
```
chshersh, line 29: Should it be `NonEmpty` list of categories? :thinking: Or is it possible not to have any categories?

src/Stan/Observation.hs
```diff
@@ -0,0 +1,32 @@
...
+import Stan.Core.Id (Id)
+import Stan.Inspection (Inspection)
+
+
+{- | Data type to represent discovered by Stan vulnerabilities.
+-}
+data Observation = Observation
+    { observationId           :: !(Id Observation)
+    , observationInspectionId :: !(Id Inspection)
+    , observationLoc          :: !SrcSpan
```
chshersh, line 27: Nice idea to store `SrcSpan` here! Do you think in the `Analysis` data type we will store something like `Map FilePath Observation` or `[(FilePath, Observation)]` for better grouping of all observations per file?

https://github.com/kowainik/stan/blob/e662cae85056ffcc0b7b353af87b0758911cccf9/src/Stan/Analysis.hs#L24

src/Stan/Inspection.hs
vrom911: Oh, nice catch. I wouldn't like to have something uncategorised. This information can be quite valuable and important for filtering and searching. What do you think?

src/Stan/Observation.hs
vrom911: Good idea. Need to think more where the filepath would be more suitable to have. But I could imagine that the tuple (or separate data type) would be more convenient here.

PR #22, opened 2020-04-12
[#6] Check for usage of 'head' for lists

### LINKED ISSUE #6: Check for usage of 'head' for lists
We should do this by analyzing HIE files and finding entrances of this function coming from `base` modules.

chshersh:
Resolves #6

It worked! On my local test case I saw this output:

```haskell
Analysis
    { analysisModuleCount = 12
    , analysisLinesOfCode = 0
    , analysisUsedExtensions = 0
    , analysisObservations =
        [ Observation
            { observationId = Id {unId = ""}
            , observationInspectionId = Id {unId = "HEAD"}
            , observationLoc = SrcSpanOneLine "app/Main.hs" 7 7 21
            , observationFile = "app/Main.hs"
            }
        ]
    }
```

### DIFF
--- a/src/Stan/Analysis.hs
+++ b/src/Stan/Analysis.hs
@@ -14,7 +14,9 @@ module Stan.Analysis
 import HieTypes (HieFile (..))
 import Relude.Extra.Lens (Lens', lens, over)
 
+import Stan.Analysis.Partial (analyseForHeadObservations)
 import Stan.Hie (countLinesOfCode)
+import Stan.Observation (Observation)
 
 
 {- | This data type stores all information collected during static analysis.
@@ -23,7 +25,7 @@ data Analysis = Analysis
     { analysisModulesNum     :: !Int
     , analysisLinesOfCode    :: !Int
     , analysisUsedExtensions :: !Int
-    , analysisObservations   :: !()  -- TODO: use Observation type later
+    , analysisObservations   :: ![Observation]
     } deriving stock (Show)
 
 modulesNumL :: Lens' Analysis Int
@@ -36,13 +38,17 @@ linesOfCodeL = lens
     analysisLinesOfCode
     (\analysis new -> analysis { analysisLinesOfCode = new })
 
+observationsL :: Lens' Analysis [Observation]
+observationsL = lens
+    analysisObservations
+    (\analysis new -> analysis { analysisObservations = new })
 
 initialAnalysis :: Analysis
 initialAnalysis = Analysis
     { analysisModulesNum     = 0
     , analysisLinesOfCode    = 0
     , analysisUsedExtensions = 0
-    , analysisObservations   = ()
+    , analysisObservations   = []
     }
 
 incModulesNum :: State Analysis ()
@@ -54,6 +60,10 @@ analised lines of code.
 incLinesOfCode :: Int -> State Analysis ()
 incLinesOfCode num = modify' $ over linesOfCodeL (+ num)
 
+-- | Add list of 'Observation's to the beginning of the existing list
+addObservations :: [Observation] -> State Analysis ()
+addObservations observations = modify' $ over observationsL (observations ++)
+
 {- | Perform static analysis of given 'HieFile'.
 -}
 runAnalysis :: [HieFile] -> Analysis
@@ -62,7 +72,8 @@ runAnalysis = executingState initialAnalysis . analyse
 analyse :: [HieFile] -> State Analysis ()
 analyse [] = pass
 analyse (hieFile:hieFiles) = do
-    incModulesNum
     -- traceM (hie_hs_file hieFile)
+    incModulesNum
     incLinesOfCode $ countLinesOfCode hieFile
+    addObservations $ analyseForHeadObservations hieFile
     analyse hieFiles
--- a/src/Stan/Analysis/Partial.hs
+++ b/src/Stan/Analysis/Partial.hs
@@ -27,7 +27,9 @@ import qualified Data.Map.Strict as Map
 -}
 analyseForHeadObservations :: HieFile -> [Observation]
 analyseForHeadObservations HieFile{..} =
-    map mkPartialObservation $ findHeads hie_asts
+    map (uncurry mkPartialObservation)
+    $ zip [1..]
+    $ findHeads hie_asts
   where
     findHeads :: HieASTs TypeIndex -> [RealSrcSpan]
     findHeads =
@@ -63,10 +65,11 @@ analyseForHeadObservations HieFile{..} =
 
         pure srcSpan
 
-    mkPartialObservation :: RealSrcSpan -> Observation
-    mkPartialObservation srcSpan = Observation
-        { observationId = Id ""  -- TODO: how to create observation id?
-        , observationInspectionId = Id "HEAD"
+    mkPartialObservation :: Int -> RealSrcSpan -> Observation
+    mkPartialObservation num srcSpan = Observation
+        -- TODO: See issue #26: https://github.com/kowainik/stan/issues/26
+        { observationId = Id $ show num <> "-STAN-0001-HEAD"
+        , observationInspectionId = Id "STAN-0001-HEAD"
         , observationLoc = srcSpan
         , observationFile = hie_hs_file
         }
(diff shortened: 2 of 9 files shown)


src/Stan/Hie/Debug.hs
```diff
@@ -36,10 +37,17 @@ import Module (Module, ModuleName, moduleNameString, moduleStableString)
...
+import Text.Pretty.Simple (pPrint)
 import Var (ArgFlag (..))
 
 import qualified Text.Show
 
+
+debugHieFile :: FilePath -> [HieFile] -> IO ()
+debugHieFile path hieFiles = do
+    let mHieFile = find (\HieFile{..} -> hie_hs_file == path) hieFiles
+    whenJust mHieFile pPrint
```
chshersh, line 49: I've decided to add `pretty-simple` to our dependencies for now, since it's useful to debug files by names while we're still learning how to work with HIE files. Can remove this later to minimize dependencies...

src/Stan/Observation.hs
```diff
@@ -13,7 +13,7 @@ module Stan.Observation
     , prettyShowObservation
     ) where
 
-import GHC (SrcSpan)
+import SrcLoc (RealSrcSpan)
```
chshersh, line 16: I had to change this since it's how it's stored in HIE files, but fortunately, it still has all necessary information

src/Stan/Observation.hs
```diff
@@ -24,7 +24,8 @@ import Stan.Inspection (Inspection)
 data Observation = Observation
     { observationId           :: !(Id Observation)
     , observationInspectionId :: !(Id Inspection)
-    , observationLoc          :: !SrcSpan
+    , observationLoc          :: !RealSrcSpan
+    , observationFile         :: !FilePath
```
chshersh, line 28: I went for the simplest solution for now and just added `FilePath` as a field. We can change this later...

src/Stan/Hie/Debug.hs
vrom911: That's okay to keep it. It will improve the debugging workflow as well :+1:

src/Stan/Analysis.hs
```diff
@@ -23,7 +25,7 @@ data Analysis = Analysis
     { analysisModuleCount    :: !Int
     , analysisLinesOfCode    :: !Int
     , analysisUsedExtensions :: !Int
-    , analysisObservations   :: !()  -- TODO: use Observation type later
+    , analysisObservations   :: ![Observation]
```
vrom911, line 28: I can imagine that in future we would like to either keep a sized list here instead or to keep a number of observations along the way. That would be handy for report + filtering/ searching. What do you think? Can keep this in mind :)

src/Stan/Analysis/Partial.hs
```diff
@@ -0,0 +1,72 @@
...
+    findHeadUsage srcSpan (identifier, _details) = do
+        Right name <- Just identifier
+
+        let occName = occNameString $ nameOccName name
+        let modul = moduleNameString $ moduleName $ nameModule name
+        let package = show @String $ moduleUnitId $ nameModule name
+
+        guard $ and
+            [ occName == "head"
+            , modul   == "GHC.List"
```
vrom911, line 60: `base` has several modules that export `head`. Are you sure that we don't need to look for each of them?

src/Stan/Analysis/Partial.hs
```diff
@@ -0,0 +1,72 @@
...
+            , package == "base"
+            ]
+
+        pure srcSpan
+
+    mkPartialObservation :: RealSrcSpan -> Observation
+    mkPartialObservation srcSpan = Observation
+        { observationId = Id ""  -- TODO: how to create observation id?
```
vrom911, line 68: That is a good question. I would like to have a number of observation plus type in the name, I think :thinking: Or maybe Module name plus something?
```
1-Module-HEAD
2-HEAD
3-Warning-HEAD
```

any of them look normal to you?

src/Stan/Analysis/Partial.hs
```diff
@@ -0,0 +1,72 @@
...
+        , observationInspectionId = Id "HEAD"
```
vrom911, line 69: I think, that having several smart constructors with the specified inspection links would make sense here.
Anyway, we have to know all of the Inspections beforehand.

src/Stan/Inspection.hs
```diff
@@ -45,3 +47,21 @@ data Severity
...
+    [ Inspection
+        { inspectionId = Id "HEAD"  -- TODO: we need to come up with naming scheme for inspection ids
+        , inspectionName = "Partial 'head'"  -- TODO: does it make sense?
+        , inspectionDescription = "Usage of partial function 'head' for lists"
+        , inspectionSolution = unlines
+            [ "* Replace list with 'NonEmpty' from 'Data.List.NonEmpty'"
+            , "* Use explicit pattern-matching over lists"
+            ]
+        , inspectionCategory =
+            one $ Category "Partial"  -- TODO: should we convert this to values to avoid typos?
```
vrom911, line 64: Yes! That would be nice, and also handy!

vrom911: We also will need a list of all categories

src/Stan/Inspection.hs
```diff
@@ -45,3 +47,21 @@ data Severity
 -- | Show 'Inspection' in a human-friendly format.
 prettyShowInspection :: Inspection -> Text
 prettyShowInspection = show
+
+{- | List of all inspections.
+-}
+inspections :: [Inspection]
+inspections =
+    [ Inspection
+        { inspectionId = Id "HEAD"  -- TODO: we need to come up with naming scheme for inspection ids
```
vrom911, line 56: Yes, another good question..
Should we use some coding scheme, like `STAN-0001-HEAD`, so we may have some other inspections in the future (like `GHC-Warning-0001`)? I am really not sure if this is helpful :thinking:

src/Stan/Inspection.hs
```diff
@@ -45,3 +47,21 @@ data Severity
...
+        , inspectionName = "Partial 'head'"  -- TODO: does it make sense?
```
vrom911, line 57: Maybe with semicolon `Partial: head`? And should we add `base/head`?

src/Stan/Analysis.hs
chshersh: @vrom911 Nice use case for the sized list! I imagine we will find more places in our codebase where we can use sized list. This data type probably will change in the future, but I agree, let's keep it in mind :slightly_smiling_face:

src/Stan/Analysis/Partial.hs
chshersh: @vrom911 Yes, I'm sure. HIE files record the original place of where the function is implemented initially.

src/Stan/Analysis/Partial.hs
chshersh: I like `2-HEAD` :slightly_smiling_face: Can implement this :+1:

src/Stan/Inspection.hs
chshersh: @vrom911 Should we move `Category` in a separate module then?

src/Stan/Inspection.hs
chshersh: I like the `STAN-0001-HEAD` scheme :heart_eyes: I agree with your reasoning that we probably will have other types of warnings, so adding `STAN` as a prefix is a good idea :+1:

src/Stan/Inspection.hs
chshersh: Excellent idea :ok_hand:

src/Stan/Inspection.hs
chshersh: Created an issue to do this separately: https://github.com/kowainik/stan/issues/25

src/Stan/Analysis/Partial.hs
vrom911: Good to know! Thanks for the explanation, @chshersh :heart:

src/Stan/Inspection.hs
vrom911: Sounds good :+1:

PR #23, opened 2020-04-12
[#21] Add test on number of lines of code

### LINKED ISSUE #21: Add test on number of lines of code

vrom911:
Resolves #21

### DIFF
--- a/test/Test/Analysis.hs
+++ b/test/Test/Analysis.hs
@@ -1,0 +1,14 @@
+module Test.Analysis
+    ( analysisSpecs
+    ) where
+
+import HieTypes (HieFile (..))
+import Test.Hspec (Spec, describe, it, shouldBe)
+
+import Stan.Hie (countLinesOfCode)
+
+
+analysisSpecs :: HieFile -> Spec
+analysisSpecs hieFile = describe "Analysis tests" $
+    it "should count lines of code of example file" $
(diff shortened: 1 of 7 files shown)


test/Test/Analysis.hs
```diff
@@ -0,0 +1,14 @@
...
+
+import HieTypes (HieFile (..))
+import Test.Hspec (Spec, describe, it, shouldBe)
+
+import Stan.Hie (countLinesOfCode)
+
+
+analysisSpecs :: HieFile -> Spec
+analysisSpecs hieFile = describe "Analysis tests" $
+    it "should count lines of code of example file" $
```
chshersh, line 13: What do you think about adding a test like `analysisModules > 10`? The number of modules will change all the time, but maybe we can at least implement some sanity check to make sure that the number of modules is calculated?

vrom911: Oh, interesting idea!
I would even prefer to have the equality check on modules, given that soon it will stabilise.
The thing is that currently there are no `.hie` file for `Spec.hs` module of tests, and I don't know the reason. This looks a bit unstable. But I think that it is already a nice sanity check — to increment the number in tests when you add another module in the project.

chshersh: This also sounds reasonable :+1: Absence of the `.hie` file for `Spec.hs` sounds like a GHC bug...

PR #27, opened 2020-04-12
[#7] Create automatic test

### LINKED ISSUE #7: Create automatic test
We can test stan on the project itself.
For this, we can create a separate stanza, where we can put Haskell modules with potentially problematic code.
The test should check the output with the expected one.

chshersh:
Resolves #7

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -36,8 +36,8 @@ jobs:
     - name: Build
       run: |
         cabal update
-        cabal build --enable-tests --enable-benchmarks
+        cabal build --enable-tests --enable-benchmarks all
 
     - name: Test
       run: |
-        cabal test --enable-tests
+        cabal test --enable-tests --test-show-details=direct all
--- a/stan.cabal
+++ b/stan.cabal
@@ -104,7 +104,7 @@ executable stan
 library test-examples
   import:              common-options
   hs-source-dirs:      target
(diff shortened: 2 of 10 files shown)


.github/workflows/ci.yml
```diff
@@ -36,8 +36,8 @@ jobs:
     - name: Build
       run: |
         cabal update
-        cabal build --enable-tests --enable-benchmarks
+        cabal build --enable-tests --enable-benchmarks all
 
     - name: Test
       run: |
-        cabal test --enable-tests
+        cabal test --enable-tests --test-show-details=direct all
```
vrom911, line 43: :rainbow:

stan.cabal
```diff
@@ -101,16 +101,25 @@ executable stan
   main-is:             Main.hs
   build-depends:       stan
 
+library test-examples
+  import:              common-options
+  hs-source-dirs:      target
+  exposed-modules:     Target.HEAD
```
vrom911, line 107: Will we have a separate file per inspection? I was thinking of having them all together :thinking: 
However, I see how testing several modules can be useful as well, but at the moment I guess one is enough. Do you ming to rename it to something more general? :slightly_smiling_face:

chshersh: @vrom911 Good idea :+1: I guess we won't ever have modules per inspection, maybe at least by some category. Renaming to something like `Target.Example` sounds reasonable for now :+1:

PR #28, opened 2020-04-13
[#8] Decide on the console output format

### LINKED ISSUE #8: Decide on the console output format
How to code warnings/errors.
What to include? (loc, filepath, module name, more?)

vrom911:
Resolves #8
image
image

### DIFF
--- a/src/Stan/Analysis/Pretty.hs
+++ b/src/Stan/Analysis/Pretty.hs
@@ -27,8 +27,11 @@ prettyShowAnalysis :: Analysis -> Text
 prettyShowAnalysis Analysis{..} = groupedObservations <> summary
   where
     groupedObservations :: Text
-    groupedObservations = Text.intercalate "\n\n" $ map showByFile $
-        toPairs $ groupObservationsByFile analysisObservations
+    groupedObservations =
+        Text.intercalate "\n\n"
+        $ map showByFile
+        $ toPairs
+        $ groupObservationsByFile analysisObservations
 
     summary :: Text
     summary = unlines
--- a/src/Stan/Analysis/Pretty.hs
+++ b/src/Stan/Analysis/Pretty.hs
@@ -37,7 +37,7 @@ prettyShowAnalysis Analysis{..} = groupedObservations <> summary
     summary = unlines
         [ ""
         , formatWith [bold] "           Stan's Summary:"
-        , separator "┏" "┳" "┓"
+        , top
         , alignText "Analysed modules" <> alignNum analysisModulesNum
         , mid
         , alignText "Analysed Lines of Code" <> alignNum analysisLinesOfCode
@@ -45,7 +45,7 @@ prettyShowAnalysis Analysis{..} = groupedObservations <> summary
         , alignText "Total extensions" <> alignNum analysisUsedExtensions
         , mid
         , alignText "Total found observations" <> alignNum (length analysisObservations)
-        , separator "┗" "┻" "┛"
+        , bot
         ]
       where
         alignNum :: Int -> Text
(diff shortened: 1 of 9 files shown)


src/Stan/Inspection.hs
```diff
@@ -48,6 +55,20 @@ data Severity
...
 
+-- | Show 'Category' in a human-friendly format.
+prettyShowCategory :: Category -> Text
+prettyShowCategory cat = formatWith [magentaBg] $ "#" <> unCategory cat
+
+-- | Get the colour of the severity level.
+severityColour :: Severity -> Text
+severityColour = \case
+    Severe -> red
+    NotReallySevere -> yellow
```
chshersh, line 66: Looks like it's time to create an issue to think about possible severities and colours for them :slightly_smiling_face:

src/Stan/Inspection.hs
```diff
@@ -48,6 +55,20 @@ data Severity
...
+
+-- | Show 'Severity' in a human-friendly format.
+prettyShowSeverity :: Severity -> Text
+prettyShowSeverity s = formatWith [severityColour s, bold] $ show s
```
chshersh, line 70: This is nice! :ok_hand:

src/Stan/Inspection.hs
```diff
@@ -57,13 +78,20 @@ inspections =
...
             one $ Category "Partial"
         , inspectionSeverity = Severe
         }
     ]
+
+-- | Get the 'Inspection' by the given inspection 'Id'.
+getInspectionById :: Id Inspection -> Inspection
+getInspectionById insId = case find ((==) insId . inspectionId) inspections of
+    Just ins -> ins
+    -- TODO: how to handle unknown ids better?
```
chshersh, line 96: It sounds like a difficult problem... Let's create an issue and discuss it? Ideally, I'd like to have a mapping from `Id Inspection` to `Inspection` where we use known `inspectionId`. But I'm not sure we can convince the compiler to do this efficiently...

src/Stan/Inspection.hs
```diff
@@ -57,13 +78,20 @@ inspections =
...
+            , "Use explicit pattern-matching over lists"
             ]
         , inspectionCategory =
             -- TODO: See issue #25: https://github.com/kowainik/stan/issues/25
             one $ Category "Partial"
         , inspectionSeverity = Severe
         }
     ]
+
+-- | Get the 'Inspection' by the given inspection 'Id'.
```
chshersh, line 92: Just to tell that the id should be known
```suggestion
-- | Get the 'Inspection' by the given known inspection 'Id'.
```

src/Stan/Observation.hs
```diff
@@ -26,8 +34,68 @@ data Observation = Observation
...
+      where
+        n :: Int
+        n = srcSpanStartLine observationLoc
+
+        alignLine :: Int -> Text
+        alignLine x = Text.justifyRight 4 ' ' (show x) <> " ┃ "
+
+        getSourceLine :: Text
+        getSourceLine = decodeUtf8 $
+            BS.lines observationFileContent !! (n - 1)
```
chshersh, line 86: I agree that the usage of partial functions here is okay. It's too much work to handle errors in here and report them, especially when they shouldn't happen..

src/Stan/Observation.hs
```diff
@@ -26,8 +34,68 @@ data Observation = Observation
...
+
+        arrows :: Text
+        arrows = severityColour (inspectionSeverity inspection)
```
chshersh, line 89: Awesome idea to use the same highlighting for arrows! :heart_eyes:

src/Stan/Analysis/Pretty.hs
```diff
@@ -0,0 +1,80 @@
...
+import Relude.Extra.Map (toPairs)
+
+import Stan.Analysis (Analysis (..))
+import Stan.Observation (Observation (..), prettyShowObservation)
+
+import qualified Data.Text as Text
+
+
+{- | Shows analysed output of Stan work.
+This functions groups 'Observation's by 'FilePath' they are found in.
```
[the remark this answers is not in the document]
vrom911: Let's discuss in here: https://github.com/kowainik/stan/issues/29

src/Stan/Inspection.hs
vrom911: Good idea: https://github.com/kowainik/stan/issues/30

src/Stan/Inspection.hs
vrom911: See https://github.com/kowainik/stan/issues/31

PR #48, opened 2020-04-19
[#47] Restructure Inspections and Analysis

### LINKED ISSUE #47: Restructure Inspections and Analysis

vrom911:
Resolves #47

### DIFF
--- a/src/Stan/Inspection/Partial.hs
+++ b/src/Stan/Inspection/Partial.hs
@@ -1,0 +1,58 @@
+{- |
+Copyright: (c) 2020 Kowainik
+SPDX-License-Identifier: MPL-2.0
+Maintainer: Kowainik <xrom.xkov@gmail.com>
+
+Contains all 'Inspection's for partial functions.
+-}
+
+module Stan.Inspection.Partial
+    ( -- * Partial 'Inspection's
+      stan0001
+    , stan0001Inspection
+
+      -- * List of all partial 'Inspection's
+    , partialInspections
+    , partialInspectionsIds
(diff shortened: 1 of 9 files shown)


src/Stan/Inspection/Partial.hs
```diff
@@ -0,0 +1,58 @@
...
+partialInspections :: [Inspection]
+partialInspections =
+    [ stan0001Inspection
+    ]
+
+-- | All partial 'Inspection's 'Id's.
+partialInspectionsIds :: [Id Inspection]
+partialInspectionsIds =
+    [ stan0001
+    ]
```
chshersh, line 34: What do you think about extracting `Id`s from the list of all inspections? We already specify the list of all inspections, so we can get all ids from this list. Or is it better to specify the list of ids explicitly as well?
```suggestion
partialInspectionsIds = map inspectionId partialInspections
```

src/Stan/Inspection/Partial.hs
```diff
@@ -0,0 +1,58 @@
...
+
+-- | Smart constructor to create partial 'Inspection'.
+mkPartialInspection :: Id Inspection -> Text -> Text -> Inspection
```
chshersh, line 37: I imagine, we'll need a data type like:

```haskell
 -- TODO: better name
data NameLocation = NameLocation
    { package :: Text
    , moduleName :: Text
    , name :: Text
    }
```

And then such smart constructors will take a value of this data type as an argument. We also will need such data type for analysis. What do you think?

Issue #40, opened 2020-04-13
Inspections for partial functions on infinite lists


vrom911:
- [x] `length`
- [x] `reverse`
- [x] `isSuffixOf`
- [x] `sum`
- [x] `product`
- [x] `genericLength`

chshersh:
There's a problem with functions `length`, `sum` and `product`. They are part of the `Foldable` typeclass, and `Prelude` exports them from `Foldable`. I imagine that in the future we will be able to check types of functions as well and we can combine multiple analysis for the same function:

1. It's either monomorphic function from the `GHC.List` module.
2. Or polymorphic function from `Data.Foldable` specialized to type `[a] -> Int`

This is not straightforward at the moment, so I propose to keep the issue open until we figure out this part.

PR #49, opened 2020-04-19
[#40] Introduce 'Infinite' inspections

chshersh:
Introduces a new category of inspections.


chshersh:
@vrom911 Good question! Unfortunately, I don't see a simpler way to define inspections at the moment... Need to think about it :thinking:

PR #59, opened 2020-04-20
[#32] Stable ids for Observations

### LINKED ISSUE #32: Stable ids for Observations
Currently, ids for `Observation`s are in the form:

```
1-STAN-0001-HEAD
```

The number in the beginning is some incremented counter. It would be better to have stable names, so people can disable specific. The idea is to include inspection id, module name and location, something like:

```
STAN-0001-HEAD-Target.Example-line7-startsOn18
```

What do you think? Is it too verbose?

chshersh:
Resolves #32

Now Observation ID looks like this:

Screenshot from 2020-04-20 20-06-13

### DIFF
--- a/src/Stan/Observation.hs
+++ b/src/Stan/Observation.hs
@@ -50,17 +50,19 @@ data Observation = Observation
 mkObservation
     :: Id Inspection  -- ^ Corresponding 'Inspection's 'Id'.
     -> HieFile
-    -> Int  -- ^ Ordinal number of the 'Observation'.
     -> RealSrcSpan  -- ^ Position.
     -> Observation
-mkObservation insId HieFile{..} num srcSpan = Observation
-    { observationId = mkObservationId num insId
+mkObservation insId HieFile{..} srcSpan = Observation
+    { observationId = mkObservationId insId modName srcSpan
     , observationInspectionId = insId
     , observationLoc = srcSpan
     , observationFile = hie_hs_file
-    , observationModuleName = toText $ moduleNameString $ moduleName hie_module
+    , observationModuleName = modName
     , observationFileContent = hie_hs_src
     }
+  where
+    modName :: Text
+    modName = toText $ moduleNameString $ moduleName hie_module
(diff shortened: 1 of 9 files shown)


src/Stan/Observation.hs
```diff
@@ -125,16 +127,22 @@ prettyShowObservation Observation{..} = unlines $
...
+1. 'Id' is stable across multiple runs of @stan@ and doesn't depend on
+other inspections in this file.
+2. 'Id' uniquely identifies 'Observation' location.
 
-The 'Observation' should look like this:
+The 'Observation' 'Id' should look like this:
 
 @
-NUM-STAN-000X-XXXX
+STAN-XXXX-Some.Module.Name-line10-column45
```
vrom911, line 139: What do you think if we add `OBS` prefix in front of all id? It would indicate clearly, that this is ID of some Observation?

vrom911: Also, I know some modules that have way to long names, it could be frustrating to look through such IDs then :slightly_frowning_face: What if we would map modules to something more short?

vrom911: Regarding"lin" and "column". I think it could be omitted. Just `11:22` could be enough, what do you think?

chshersh: @vrom911 These are all excellent points!

test/Test/Stan/Observation.hs
```diff
@@ -0,0 +1,24 @@
...
+
+import Stan.Core.Id (Id (..))
+import Stan.Inspection.Partial (stan0001)
+import Stan.Observation (Observation, mkObservationId)
+
+
+observationSpec :: Spec
+observationSpec = describe "Observation" $
+    it "calculates Observation Id properly" $
+        testObservationId `shouldBe` Id "OBS-STAN-0001-bbjjJQ-10:42"
```
vrom911, line 16: Nice tests :heart_eyes_cat:

PR #60, opened 2020-04-20
[#31] Store inspections in 'HashMap'

### LINKED ISSUE #31: Store inspections in 'HashMap'
See https://github.com/kowainik/stan/pull/28#discussion_r407524603

vrom911:
Resolves #31

### DIFF
--- a/src/Stan/Core/Id.hs
+++ b/src/Stan/Core/Id.hs
@@ -2,7 +2,6 @@
 
 {-# LANGUAGE AllowAmbiguousTypes #-}
 {-# LANGUAGE DataKinds           #-}
-{-# LANGUAGE DeriveAnyClass      #-}
 {-# LANGUAGE TypeFamilies        #-}
 {-# LANGUAGE TypeOperators       #-}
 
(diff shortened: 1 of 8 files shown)


src/Stan/Core/Id.hs
```diff
@@ -27,8 +28,9 @@ structures by using a phantom parameter.
 -}
 newtype Id a = Id
     { unId :: Text
-    } deriving stock (Show)
+    } deriving stock (Show, Generic)
       deriving newtype (Eq)
+      deriving anyclass (Hashable)
```
chshersh, line 33: I think we can derive `Hashable` using the `newtype` strategy here, since `Text` already has `Hashable` instance

src/Stan/Inspection/Infinite.hs
```diff
@@ -22,29 +22,24 @@ module Stan.Inspection.Infinite
...
 import Stan.NameMeta (NameMeta (..), mkBaseListMeta)
 import Stan.Severity (Severity (..))
 
 
--- | All infinite 'Inspection's.
-infiniteInspections :: [Inspection]
-infiniteInspections =
+-- | All infinite 'Inspection's map from 'Id's.
+infiniteInspectionsMap :: InspectionsMap
+infiniteInspectionsMap = fromList $ map (\i -> (inspectionId i, i))
```
chshersh, line 37: Looks like `\i -> (inspectionId i, i)` can be replaced by [`mapToFst`](http://hackage.haskell.org/package/relude-0.6.0.0/docs/Relude-Extra-Tuple.html#v:mapToFst). But we have an open issue in `relude` regarding better naming for this function...

PR #63, opened 2020-04-21
[#34] Group observations into 'Slist'

### LINKED ISSUE #34: Group observations into 'Slist'
Currently, `Observations` are displayed in non-uniform order. For better consistency, I propose to sort them by source location.

Also, while we're on this, let's think about the data structure for all `Observation`s. Currently, it's a list. But what we want is:

1. Calculate length fastly.
2. Append different pieces from different inspections.

Something like `slist` should work nice here, as @vrom911 suggested. But we probably will need to implement custom grouping function:

```haskell
groupBy :: (a -> k) -> [a] -> Map k (Slist a)
```

vrom911:
Resolves #34

### DIFF
--- a/src/Stan/Analysis/Pretty.hs
+++ b/src/Stan/Analysis/Pretty.hs
@@ -17,7 +17,7 @@ import Relude.Extra.Map (toPairs)
 import Stan.Analysis (Analysis (..))
 import Stan.Core.ModuleName (ModuleName (..))
 import Stan.Inspection.All (inspectionsMap)
-import Stan.Observation (Observation (..), prettyShowObservation)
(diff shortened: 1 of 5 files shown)


src/Stan/Analysis/Pretty.hs
```diff
@@ -81,6 +81,6 @@ showByFile (file, o :| obs) = unlines
     b = formatWith [bold]
 
 groupObservationsByFile
-    :: [Observation]
+    :: Observations
     -> HashMap FilePath (NonEmpty Observation)
```
chshersh, line 85: Do you think it makes sense to change the type here to `HashMap FilePath Observations` so we can tell the total number of observations in a group faster?

PR #64, opened 2020-04-21
[#61] Simplify code structure for Inspection

### LINKED ISSUE #61: Simplify code structure for Inspection
Proposal to change the current code to the following scheme:

1. Remove old `stanXXXX :: Id Inspection` variable.
2. Rename `stanXXXInspection :: Inspection` to simply `stanXXXX :: Inspection`. We have the `inspectionId :: Inspection -> Id Inspection` function to get the id.
3. Remove `stanXXXXMeta` variables.
4. Remove `stanXXXXAnalyser` variables.
5. Add to `Inspection` a field of some big sum type that represents different kinds of our inspections. I'm thinking about something like this:

```haskell
data InspectionAnalysis
    = FindName NameMeta  -- ^ For searching for  a specific functions
    | Infix
    ...
```
6. Create a function: `analysisByInspection :: InspectionAnalysis -> HieFile -> [Observation]`
7. Refactor the `analyse` function to the following scheme: traverse list of all inspections run the corresponding analysis function for the `InspectionAnalysis` constructor.

With this code structure, we should be able to reduce the number of variables and documentation we need to maintain.

What do you think?

chshersh:
Resolves #61

Mostly code removal and tests speedup :rofl:


src/Stan/Inspection.hs
```diff
@@ -29,13 +31,24 @@ data Inspection = Inspection
...
 type InspectionsMap = HashMap (Id Inspection) Inspection
 
+{- | Data type that represents all possible types of @stan@
+inspections in a uniformed way.
+-}
+data InspectionAnalysis
+    -- | Find specific function name.
+    = FindName NameMeta
+    -- | Missing @infix@ declaration for operator.
+    | Infix
```
chshersh, line 49: Better naming conventions are always welcome :slightly_smiling_face:

src/Stan/Analysis/Analyser.hs
```diff
@@ -30,19 +30,13 @@ import qualified Data.Set as Set
...
 
 
-{- | Contains analyser function to run on 'HieFile's.
+{- | Create analysing function for 'Inspection' by pattern-matching
+over 'InspectionAnalysis'.
 -}
-data Analyser = Analyser
-    { analyserInspectionId :: !(Id Inspection)
-    , analyserFunction     :: !(HieFile -> Observations)
-    }
```
chshersh, line 38: For now, there is no need in the `Analyser` data type, but we may want to return it in the future when we need it.

src/Stan/Analysis/Analyser.hs
```diff
@@ -30,19 +30,13 @@ import qualified Data.Set as Set
...
--- | Smart constructor for 'Analyser' creation of partial functions.
-mkNameMetaAnalyser :: Id Inspection -> NameMeta -> Analyser
-mkNameMetaAnalyser insId nameMeta = Analyser
-    { analyserInspectionId = insId
-    , analyserFunction = analyseNameMeta insId nameMeta
-    }
+analysisByInspection :: Inspection -> HieFile -> Observations
+analysisByInspection Inspection{..} = case inspectionAnalysis of
+    FindName nameMeta -> analyseNameMeta inspectionId nameMeta
+    Infix             -> const mempty  -- TODO: not yet implemented
```
chshersh, line 39: I've created two constructors, so HLint won't suggest to use `newtype` instead of `data`

Issue #38, opened 2020-04-13
Inspections for partial functions


vrom911:
### Lists

- [x] `head`
- [x] `tail`
- [x] `init`
- [x] `last`
- [x] `(!!)`
- [x] `fromJust`
- [x] `maximum`
- [x] `minimum`
- [x] `maximumBy`
- [x] `minimumBy`
- [x] `foldl1`
- [x] `foldl1'`
- [x] `foldr1`
- [x] `cycle`
- [x] `genericIndex`

### Other

- [x] `read`
- [x] `succ`
- [x] `pred`
- [x] `toEnum`

PR #68, opened 2020-04-23
[#38] Add 'genericIndex' partial inspection

chshersh:


test/Test/Stan/Analysis/Common.hs
```diff
@@ -28,7 +30,7 @@ observationAssert modulePath moduleName analysis Inspection{..} line start end =
   where
     foundPartialObservation :: Maybe Observation
     foundPartialObservation = find
-        (\Observation{..} -> observationId == obsId)
+        (\Observation{..} -> obsIdShort `Text.isPrefixOf` unId observationId)
```
vrom911, line 33: Nice trick :+1:

PR #71, opened 2020-04-25
[#33] Add --hide-solution option

### LINKED ISSUE #33: Add --solution option in CLI to hide/show solution
Some would like to have their output short and omit hints/solutions in there.

vrom911:
Resolves #33

### DIFF
--- a/src/Stan/Core/Toggle.hs
+++ b/src/Stan/Core/Toggle.hs
@@ -1,0 +1,25 @@
+{- |
+Copyright: (c) 2020 Kowainik
+SPDX-License-Identifier: MPL-2.0
+Maintainer: Kowainik <xrom.xkov@gmail.com>
+
+Data types to show/hide different parts of inspections and observations.
+-}
+
+module Stan.Core.Toggle
+    ( ToggleSolution (..)
+
+    , isHidden
+    ) where
+
+
+-- | Boolean for showing/hiding solution information of observations.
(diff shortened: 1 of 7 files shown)


src/Stan/Core/Toggle.hs
```diff
@@ -0,0 +1,25 @@
+{- |
+Copyright: (c) 2020 Kowainik
+SPDX-License-Identifier: MPL-2.0
+Maintainer: Kowainik <xrom.xkov@gmail.com>
+
+Data types to show/hide different parts of inspections and observations.
```
chshersh, line 6: This is a very nice module! I can see how it can be even more useful in the future :+1:

src/Stan/Observation.hs
```diff
@@ -129,10 +130,13 @@ prettyShowObservation Observation{..} = unlines $
             arrow = srcSpanEndCol observationLoc - start - 1
 
     solution :: [Text]
-    solution = case inspectionSolution inspection of
-        []   -> []
-        sols -> formatWith [italic, green] "Possible solution:" :
+    solution
+        | isHidden toggleSolution = []
+        | null sols = []
```
chshersh, line 135: Do you think it makes sense to simplify condition or it's better to specify it on separate lines for better visibility?
```suggestion
        | isHidden toggleSolution || null sols = []
```

PR #79, opened 2020-04-29
[#75] Find filepath to cabal file

### LINKED ISSUE #75: Find filepath to cabal file
This is needed for extensions information calculations.

vrom911:
Resolves #75

### DIFF
--- a/src/Stan.hs
+++ b/src/Stan.hs
@@ -10,9 +10,11 @@ module Stan
     ( run
     ) where
 
-import Colourista (errorMessage, formatWith, italic)
+import Colourista (errorMessage, formatWith, infoMessage, italic, warningMessage)
 import Extensions (getPackageExtentionsBySources)
 import HieTypes (HieFile (..))
+import System.Directory (getCurrentDirectory, listDirectory)
+import System.FilePath (takeExtension, (</>))
 
 import Stan.Analysis (runAnalysis)
 import Stan.Analysis.Pretty (prettyShowAnalysis)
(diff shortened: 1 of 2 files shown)


src/Stan.hs
```diff
@@ -35,7 +37,15 @@ runStan StanArgs{..} = do
     -- create Map from filepathes to sources.
     let sourcesMap = fromList $ map (\HieFile{..} -> (hie_hs_file, hie_hs_src)) hieFiles
     -- create extensions map
-    extensionsMap <- getPackageExtentionsBySources "stan.cabal" sourcesMap
+    extensionsMap <- findCabalFile >>= \case
+        Just cabal -> do
+            infoMessage $ "Using the following .cabal file: " <> toText cabal <> "\n"
+            getPackageExtentionsBySources cabal sourcesMap
+        -- if cabal file is not found, pass the empty map instead
```
chshersh, line 44: Good solution for now 👍  I think in the future we can support work even without the `.cabal` file by collecting extensions only in modules. So we somehow need to parse each module separately, if the `.cabal` file not found. I don't see a straight solution at the moment, but should be possible.  Like, with printing the warning that cabal file not found and that the list of extensions might not be complete.

src/Stan.hs
```diff
@@ -35,7 +37,15 @@ runStan StanArgs{..} = do
...
+        Nothing    -> do
+            warningMessage ".cabal file not found in the current directory."
+            infoMessage " 💡 Try using --cabal-file option to specify the path to the .cabal file.\n"
```
chshersh, line 47: Excellent 💡 to use this emoji! ❤️

src/Stan.hs
vrom911: Excellent point! I've created the issue for that: https://github.com/kowainik/stan/issues/82

PR #84, opened 2020-04-30
[#58] Implement type search in HIE files

### LINKED ISSUE #58: Implement type search in HIE files
When working on some inspections, it was discovered that we need to implement some sort of search in types or matching of types of some expressions. Fortunately, HIE files contain information about types. Unfortunately, they contain information in _compact_ way which is not always easy to process. So if you have a type like this:

```haskell
foo :: NonEmpty String -> Int
```

It will be stored as an `Array` like this (I write as a mapping, but index is implicit in arrays):

```haskell
1 -> "Int"      []
2 -> "String"   []
3 -> "NonEmpty" [ 2 ]
4 -> FunType    3 1
```

We need to implement the following:

1. Interface to specify search queries. We need to think about type-safe and convenient way to specify queries via some nice eDSL. Sometimes we need to search for exact types, like `fromInteger :: Natural -> Integer` and sometimes we don't care about all type parameters and want to have some sort of wildcards, like `length :: [_] -> Int`.
2. Matching function that will search the array of types recursively.

When verifying the type of the expression, we know the index of that type (`Int`, which is called `TypeIndex` in the `HieTypes` module). So a matching function will have the following shape:

```haskell
matchType
    :: TypeIndex   -- ^ Index of the current expression type
    -> TypePattern  -- ^ Our search query
    -> Array TypeIndex HieTypeFlat  -- ^ Array of all types in HIE file
    -> Maybe HieTypeFlat  -- ^ Matched type, if it was found
```

vrom911:
Resolves #58

### DIFF
--- a/src/Stan/Hie/Match.hs
+++ b/src/Stan/Hie/Match.hs
@@ -1,0 +1,95 @@
+{- |
+Copyright: (c) 2020 Kowainik
+SPDX-License-Identifier: MPL-2.0
+Maintainer: Kowainik <xrom.xkov@gmail.com>
+
+Some `Stan.Inspection.Inspection`s require to know about types and some
+mechanism to match types to the given 'Pattern'. This information on types/type
+expressions is taken from @HIE files@ in a more suitable view.
+
+Let's take a look at the function @foo@:
+
+@
+foo :: NonEmpty String -> Int
+@
+
+In @HIE@ files it will be stored as an 'Array' like this:
+
+@
+  1 -> "Int"      []
+  2 -> "String"   []
+  3 -> "NonEmpty" [ 2 ]
+  4 -> FunType    3 1
+@
+
+This module contains an implementation of the process of retrieval of this
+information from there.
+-}
+
+module Stan.Hie.Match
+    ( Pattern (..)
+    , hieMatchType
+    ) where
+
+import BasicTypes (PromotionFlag (NotPromoted))
+import Data.Array (Array)
+import HieTypes (HieArgs (..), HieType (..), HieTypeFlat, TypeIndex)
+import IfaceType (IfaceTyCon (..), IfaceTyConInfo (..))
+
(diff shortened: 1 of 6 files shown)


src/Stan/Hie/Match.hs
```diff
@@ -0,0 +1,95 @@
...
+    {- | Argument, type or constructor:
+
+    +---------------------+---------------------------------------------------------------------+
+    | @a@                 | @PatternName (NameMeta ... \"a\") []@                               |
+    +---------------------+---------------------------------------------------------------------+
+    | @[a]@               | @PatternName (NameMeta ... \"[]\") [aPattern]@                      |
+    +---------------------+---------------------------------------------------------------------+
+    | @Either Int String@ | @PatternName (NameMeta ... \"Either\") [intPattern, stringPattern]@ |
+    +---------------------+---------------------------------------------------------------------+
+    -}
```
chshersh, line 56: This documentation is super amazing! :heart:

src/Stan/Hie/Match.hs
```diff
@@ -0,0 +1,95 @@
...
+    then Just curFlat
+    else Nothing
+  where
+    curFlat :: HieTypeFlat
+    curFlat = arr Arr.! i
+
+    satisfyPattern :: HieTypeFlat -> Pattern -> Bool
+    satisfyPattern (HTyVarTy name) (PatternName nameMeta []) =
+        compareNames nameMeta name
+    satisfyPattern (HTyConApp IfaceTyCon{..} (HieArgs hieArgs)) (PatternName nameMeta args) =
```
chshersh, line 81: Do you think it's worth it to split this long line into multiple lines? Or it's not too long?
```suggestion
    satisfyPattern
        (HTyConApp IfaceTyCon{..} (HieArgs hieArgs))
        (PatternName nameMeta args)
      =
```

PR #85, opened 2020-05-01
[#40] Inspections for partial functions on infinite lists

### LINKED ISSUE #40: Inspections for partial functions on infinite lists

vrom911:
Resolves #40

### DIFF
--- a/src/Stan/Analysis/Analyser.hs
+++ b/src/Stan/Analysis/Analyser.hs
@@ -17,6 +17,7 @@ import Slist (Slist, slist)
 import SrcLoc (RealSrcSpan)
 
 import Stan.Core.Id (Id)
+import Stan.Hie.Match (Pattern (..), hieMatchType)
 import Stan.Inspection (Inspection (..), InspectionAnalysis (..))
 import Stan.NameMeta (NameMeta (..), compareNames)
 import Stan.Observation (Observations, mkObservation)
@@ -31,21 +32,22 @@ over 'InspectionAnalysis'.
 -}
 analysisByInspection :: Inspection -> HieFile -> Observations
 analysisByInspection Inspection{..} = case inspectionAnalysis of
-    FindName nameMeta -> analyseNameMeta inspectionId nameMeta
-    Infix             -> const mempty  -- TODO: not yet implemented
+    FindName nameMeta pat -> analyseNameMeta inspectionId nameMeta pat
+    Infix                 -> const mempty  -- TODO: not yet implemented
 
 {- | Check for occurrences of the specified function given via 'NameMeta'.
 -}
 analyseNameMeta
   :: Id Inspection
   -> NameMeta
+  -> Pattern
   -> HieFile
   -> Observations
-analyseNameMeta insId nameMeta hie@HieFile{..} =
-    mkObservation insId hie <$> findHeads hie_asts
+analyseNameMeta insId nameMeta pat hie@HieFile{..} =
+    mkObservation insId hie <$> findSpans hie_asts
   where
-    findHeads :: HieASTs TypeIndex -> Slist RealSrcSpan
-    findHeads =
+    findSpans :: HieASTs TypeIndex -> Slist RealSrcSpan
+    findSpans =
         S.concatMap findInAst
         . Map.elems
         . getAsts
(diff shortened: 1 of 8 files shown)


src/Stan/Analysis/Analyser.hs
```diff
@@ -56,17 +58,20 @@ analyseNameMeta insId nameMeta hie@HieFile{..} =
...
         -> (Identifier, IdentifierDetails TypeIndex)
         -> Maybe RealSrcSpan
-    findHeadUsage srcSpan (identifier, details) = do
+    findUsage typeIxs srcSpan (identifier, details) = do
         Right name <- Just identifier
         guard $ Set.notMember (IEThing Import) $ identInfo details
 
         guard $ compareNames nameMeta name
 
+        guard $ any isJust $ map (\i -> hieMatchType i pat hie_types) typeIxs
```
chshersh, line 75: Would it be better to have a single `guard` with multiple cases? Like:

```haskell
guard
    $  Set.nonMember ...
    && compareNames ...
    && ...
```

src/Stan/Inspection/Infinite.hs
```diff
@@ -35,29 +42,44 @@ infiniteInspectionsMap = fromList $ map (mapToFst inspectionId)
...
     , inspectionCategory = Category.infinite :| [Category.list]
     , inspectionSeverity = Warning
-    , inspectionAnalysis = FindName nameMeta
+    , inspectionAnalysis = FindName nameMeta pat
     }
 
 -- | 'Inspection' for 'stan0101' — infinite 'reverse' @STAN-0101@.
 stan0101 :: Inspection
-stan0101 = mkInfiniteInspection (Id "STAN-0101") (mkBaseListMeta "reverse")
+stan0101 = mkInfiniteInspection (Id "STAN-0101") (mkBaseListMeta "reverse") PatternAnything
```
chshersh, line 65: In such cases I with Haskell supported default arguments...

src/Stan/Inspection/Infinite.hs
```diff
@@ -35,29 +42,44 @@ infiniteInspectionsMap = fromList $ map (mapToFst inspectionId)
...
+stan0104 :: Inspection
+stan0104 = mkInfiniteInspection (Id "STAN-0104") (mkBaseOldListMeta "genericLength") PatternAnything
+
+-- | 'Inspection' for 'stan0105' — infinite 'Data.Foldable.sum' @STAN-0105@.
+stan0105 :: Inspection
+stan0105 = mkInfiniteInspection (Id "STAN-0105") (mkBaseFoldableMeta "sum") (PatternFun patternList PatternAnything)
+
+-- | 'Inspection' for 'stan0106' — infinite 'Data.Foldable.product' @STAN-0106@.
+stan0106 :: Inspection
+stan0106 = mkInfiniteInspection (Id "STAN-0106") (mkBaseFoldableMeta "product") (PatternFun patternList PatternAnything)
```
chshersh, line 85: This pattern seems common enough. Should we create a variable for it, like `listFunPattern`?

src/Stan/Inspection/Infinite.hs
vrom911: :crying_cat_face:

src/Stan/Hie/Match.hs
```diff
@@ -66,15 +66,12 @@ data Pattern
 
 {- | Matching function that searches the array of types recursively.
 -}
-hieMatchType
+hieMatchPattern
     :: TypeIndex   -- ^ Index of the current expression type
     -> Pattern  -- ^ Our search query
     -> Array TypeIndex HieTypeFlat  -- ^ Array of all types in HIE file
```
chshersh, line 72: Here is just a thought I had: what if we reorder arguments in this function to the following order:

```haskell
hieMatchPattern :: Array TypeIndex -> Pattern -> TypeIndex -> Bool
```

With the new order you whould be able to use partially applied `hieMatchPattern` and rewrite things like `any (\i -> hieMatchPattern i pat hie_types)` to `any (hieMatchPattern hie_types pat)`. Also, probably, with some extra internal functions, you can avoid passing `arr` as an argument inside internal recursive functions, but that's a minor suggestion, maybe not worth it.

src/Stan/Hie/Match.hs
```diff
@@ -89,12 +86,12 @@ hieMatchType i pat arr =
...
+        && checkWith (\(_, ix) a -> hieMatchPattern ix a arr) hieArgs args
     satisfyPattern (HFunTy i1 i2) (PatternFun p1 p2) =
-           isJust (hieMatchType i1 p1 arr)
-        && isJust (hieMatchType i2 p2 arr)
-    satisfyPattern (HQualTy _ ix) p = isJust $ hieMatchType ix p arr
-    satisfyPattern (HForAllTy _ ix) p = isJust $ hieMatchType ix p arr
+           hieMatchPattern i1 p1 arr
+        && hieMatchPattern i2 p2 arr
+    satisfyPattern (HQualTy _ ix) p = hieMatchPattern ix p arr
+    satisfyPattern (HForAllTy _ ix) p = hieMatchPattern ix p arr
```
chshersh: I like the `Bool` simplification! Made code much simpler 👍

src/Stan/Analysis/Analyser.hs
```diff
@@ -56,17 +58,22 @@ analyseNameMeta insId nameMeta hie@HieFile{..} =
...
-        guard $ Set.notMember (IEThing Import) $ identInfo details
-
-        guard $ compareNames nameMeta name
+        guard
+            -- not in the imports
+            $ Set.notMember (IEThing Import) (identInfo details)
+            -- exact name/module/package
+            && compareNames nameMeta name
+            -- compatible with the given pattern
+            && any (hieMatchPattern hie_types pat) typeIxs
```
chshersh, line 77: This is genius! 😍

PR #92, opened 2020-05-01
[#90] Automate inspection testing

### LINKED ISSUE #90: Automate inspection testing

vrom911:
Resolves #90


target/Target/Partial.hs
```diff
@@ -7,6 +7,8 @@ import Data.List.NonEmpty (NonEmpty)
 import Data.Maybe (fromJust)
 import GHC.Exts (fromList)
 import Numeric.Natural (Natural)
+
```
chshersh, line 10: Should we also add an empty line after pragma?

PR #93, opened 2020-05-01
[#76] Store file map in Analysis

### LINKED ISSUE #76: Store file map in Analysis
This will help to keep per file information that could be used for more detailed reports.
For example, it could contain LoC, extensions per file.

### LINKED ISSUE #82: When cabal file is not found get extensions per module instead of none

vrom911:
Resolves #76
Resolves #82

### DIFF
--- a/src/Stan.hs
+++ b/src/Stan.hs
@@ -10,8 +10,12 @@ module Stan
     ( run
     ) where
 
+import Relude.Extra.Tuple (mapToSnd)
+
 import Colourista (errorMessage, formatWith, infoMessage, italic, warningMessage)
-import Extensions (ExtensionsResult, getPackageExtentionsBySources)
+import Control.Exception (catch)
+import Extensions (ExtensionsError (CabalParseError), ExtensionsResult)
+import Extensions.Cabal (CabalException, parseCabalFileExtensions)
 import HieTypes (HieFile (..))
 import System.Directory (doesFileExist, getCurrentDirectory, listDirectory)
 import System.FilePath (takeExtension, (</>))
@@ -25,6 +29,9 @@ import Stan.Inspection (prettyShowInspection, prettyShowInspectionShort)
 import Stan.Inspection.All (inspections, lookupInspectionById)
 -- import Stan.Hie.Debug (debugHieFile)
 
+import qualified Data.Map.Strict as Map
+import qualified Data.Set as Set
+
 
 run :: IO ()
 run = runStanCli >>= \case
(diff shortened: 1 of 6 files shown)


src/Stan.hs
```diff
@@ -34,8 +41,8 @@ run = runStanCli >>= \case
 runStan :: StanArgs -> IO ()
 runStan StanArgs{..} = do
     hieFiles <- readHieFiles stanArgsHiedir
-    -- create extensions map
-    extensionsMap <- case stanArgsCabalFilePath of
+    -- create cabal default extensions map
+    cabalExtensionsMap <- case stanArgsCabalFilePath of
```
chshersh, line 45: Would it make sense to move all code related to creating `extensionsMap` into a separate function, so the main function will look simpler? Like:

```haskell
runStan StanArgs{..} = do
    hieFiles <- readHieFiles
    cabaExtensionsMap <- createCabalExtensionsMap stanArgsCabalFilePath hieFiles
    let analysis = ...
    putTextLn ...
```

It's not necessary, but I wonder whether it will improve code readability :thinking:

src/Stan.hs
```diff
@@ -50,17 +57,21 @@ runStan StanArgs{..} = do
...
         infoMessage $ "Using the following .cabal file: " <> toText cabal <> "\n"
-        getPackageExtentionsBySources cabal sourcesMap
+        (Right . Set.fromList <<$>> parseCabalFileExtensions cabal)
+            `catch` handleCabalErr
+      where
+        handleCabalErr :: CabalException -> IO (Map FilePath ExtensionsResult)
+        handleCabalErr _ = do
+            errorMessage "Error when parsing cabal file. Stan will continue without information from .cabal file"
+            pure $ Map.fromList $
+                map (mapToSnd (const $ Left CabalParseError) . hie_hs_file) hies
```
chshersh, line 74: As we discussed, it should be possible to add a constructor to the `ExtensionError` type to store `CabalParseError`, so we can fill the map with the specific `CabalException` we get in this function. Do you think it should be done in the `extensions` library?

src/Stan/FileInfo.hs
```diff
@@ -0,0 +1,27 @@
...
+import Extensions.OnOff (OnOffExtension)
+
+
+-- | File specific information.
+data FileInfo = FileInfo
+    { fileInfoPath            :: !FilePath
+    , fileInfoLoc             :: !Int
+    , fileInfoCabalExtensions :: !ExtensionsResult
+    , fileInfoExtensions      :: !ExtensionsResult
+    , fileMergedExtensions    :: !(Set OnOffExtension)
```
chshersh, line 24: I guess in the future we can also add `Slist Observation` to this data type as well to get grouping by `FilePath` for free.

src/Stan.hs
vrom911: Yes, we can improve the experience in there as well :+1:

PR #101, opened 2020-05-03
[#96] Use latest version of extensions, support SafeHaskellExtensions

### LINKED ISSUE #96: Use latest version of extensions, support SafeHaskellExtensions

vrom911:
Resolves #96

### DIFF
--- a/src/Stan/Analysis.hs
+++ b/src/Stan/Analysis.hs
@@ -11,9 +11,8 @@ module Stan.Analysis
     , runAnalysis
     ) where
 
-import Extensions (ExtensionsError (ModuleParseError, NotCabalModule), ExtensionsResult)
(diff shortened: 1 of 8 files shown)


src/Stan/Analysis.hs
```diff
@@ -67,7 +66,7 @@ initialAnalysis :: Analysis
 initialAnalysis = Analysis
     { analysisModulesNum     = 0
     , analysisLinesOfCode    = 0
-    , analysisUsedExtensions = mempty
+    , analysisUsedExtensions = (mempty, mempty)
```
chshersh, line 69: This actually still can be just `mempty` :smirk: 
```suggestion
    , analysisUsedExtensions = mempty
```

PR #103, opened 2020-05-04
[#97] eDSL for 'Pattern'

### LINKED ISSUE #97: eDSL for 'Pattern'
We start to define more and more type patterns, so it could be beneficial to create a module like `Stan.Pattern` with the data type, common patterns and some eDSL to define patterns easily. A few ideas:

* `PatternFun` is `~>` or `:>` or `:->` or something like that
* `PatternAnything` can be nullary operator `(?)` or a name like `_Any`
* `PatternOr` can be `|||`
* `PatternName` — :thinking: can't come up with anything small

Any other ideas are appreciated :slightly_smiling_face:

chshersh:
Resolves #97

Let me know what do you think and whether there're better ideas :slightly_smiling_face:

### DIFF
--- a/src/Stan/Pattern.hs
+++ b/src/Stan/Pattern.hs
@@ -1,0 +1,104 @@
+{- |
+Copyright: (c) 2020 Kowainik
+SPDX-License-Identifier: MPL-2.0
+Maintainer: Kowainik <xrom.xkov@gmail.com>
+
+Patterns for types and type search.
+-}
+
+module Stan.Pattern
+    ( -- * Type
+      Pattern (..)
+
+      -- * eDSL
+    , (?)
+    , (|||)
+    , (|->)
+    , (|::)
+
+      -- * Common 'Pattern's
+    , listPattern
+    , nonEmptyPattern
+    , listFunPattern
+    , integerPattern
+    , naturalPattern
+    ) where
+
+import Relude.Extra.Lens ((.~))
+
+import Stan.NameMeta (NameMeta (..), mkBaseMeta, moduleNameL)
+
+
+{- | Query pattern used to search types in HIE AST.
+-}
+data Pattern
+    {- | Argument, type or constructor:
+
+    +---------------------+---------------------------------------------------------------------+
+    | @a@                 | @PatternName (NameMeta ... \"a\") []@                               |
+    +---------------------+---------------------------------------------------------------------+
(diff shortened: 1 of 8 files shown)


src/Stan/Pattern.hs
```diff
@@ -0,0 +1,104 @@
...
+integerPattern = NameMeta
+    { nameMetaName       = "Integer"
+    , nameMetaModuleName = "GHC.Integer.Type"
+    , nameMetaPackage    = "integer-wired-in"
+    } |:: []
+
+-- | 'Pattern' for 'Natural'.
+naturalPattern :: Pattern
+naturalPattern =
+    (mkBaseMeta "Natural" & moduleNameL .~ "GHC.Natural") |:: []
```
chshersh, line 104: I wanted some short name for this, so I could write something like:

```suggestion
    baseNameMeta "Natural" "GHC.Natural" |:: []
```

or even

```haskell
    "Natural" `baseNameFrom` "GHC.Natural" |:: []
```

Do you think it something we should add?

src/Stan/Pattern.hs
```diff
@@ -0,0 +1,104 @@
...
+-- | 'Pattern' for list @[a]@ or @'String'@.
+listPattern :: Pattern
+listPattern =
+    NameMeta
+        { nameMetaName       = "[]"
+        , nameMetaModuleName = "GHC.Types"
+        , nameMetaPackage    = "ghc-prim"
+        } |:: [ (?) ]
+    |||
+    (mkBaseMeta "String" & moduleNameL .~ "GHC.Base") |:: []
```
chshersh, line 82: This may look like operator overdose, but I guess the main goal of implementing eDSL is to be able to write shorter code. A few extra ideas we can do:

1. Move big `NameMeta`s into variables, so it could be just `listNameMeta |:: [ (?) ] ||| ...`
2. Create one more operator for cases where we have an only single type, like
    ```haskell
    listName |:: (?)
   ```

vrom911: I think I like a first option more (for consistency) 🤔

src/Stan/Pattern.hs
vrom911:
```haskell
"Natural" `baseNameFrom` "GHC.Natural" |:: []
```

^^ This is 🔝

Issue #65, opened 2020-04-22
Find occurrences of [0 .. length xs]


chshersh:
This expression is represented in the HIE file in the following way (shortened for a clearer picture):

```haskell
Node
    { annotations = [ ("ArithSeq", "HsExpr") ]
    , children =
        [ Node
            { annotations = [ ( "HsOverLit", "HsExpr") ]
            , children = []
            }
        , Node
            { annotations = [ ( "HsApp", "HsExpr") ]
            , children = [ length, xs ]
            }
        ]
    }
```

So, the algorithm to find a piece of the AST is the following:

1. Traverse AST until we find the `("ArithSeq", "HsExpr")` pair in the node context. This means that this node represents a range of the shape `[a .. b]`.
2. Pattern-match on children of this AST and verify they have the desired shape:
    * First children is "HsOverLit" equal to `0` (equality check is a bit tricky, since the constant is not stored in AST, only it's position in the source file)
    * Second children is a function application of the "length" function to anything

### Full AST

```haskell
Node
  { nodeInfo = NodeInfo
      { nodeAnnotations = fromList
          [
              ( "ArithSeq"
              , "HsExpr"
              )
          ]
      , nodeType = []
      , nodeIdentifiers = fromList []
      }
  , nodeSpan = SrcSpanOneLine "target/Target/Infinite.hs" 10 12 28
  , nodeChildren =
      [ Node
          { nodeInfo = NodeInfo
              { nodeAnnotations = fromList
                  [
                      ( "HsOverLit"
                      , "HsExpr"
                      )
                  ]
              , nodeType = []
              , nodeIdentifiers = fromList []
              }
          , nodeSpan = SrcSpanOneLine "target/Target/Infinite.hs" 10 13 14
          , nodeChildren = []
          }
      , Node
          { nodeInfo = NodeInfo
              { nodeAnnotations = fromList
                  [
                      ( "HsApp"
                      , "HsExpr"
                      )
                  ]
              , nodeType = []
              , nodeIdentifiers = fromList []
              }
          , nodeSpan = SrcSpanOneLine "target/Target/Infinite.hs" 10 18 27
          , nodeChildren =
              [ Node
                  { nodeInfo = NodeInfo
                      { nodeAnnotations = fromList
                          [
                              ( "HsVar"
                              , "HsExpr"
                              )
                          ,
                              ( "HsWrap"
                              , "HsExpr"
                              )
                          ]
                      , nodeType =
                          [ 7
                          , 18
                          ]
[...]

PR #113, opened 2020-05-21
[#107] Patterns for AST

### LINKED ISSUE #107: Patterns for AST
We need to pattern-match not only over types but also over AST. So we need to define a data type for simple and convenient ways of defining AST patterns and also for matching over those patterns. Currently we have 4 AST patterns:

* `[0 .. length xs]`
* `for_ [n ... m]` or `forM_ [n .. m]`
* `infix n {operator_1}, {operator_2} ... {operator_M}`
* `function definition` (for tracking operator definitions and counting the number of defined functions)
* `{possible_string_url} </> ...`

This is a bit trickier than patterns for types because AST in HIE is not in the convenient format. As described in #65, AST is represented as a tree of nodes with string tags. For example, AST for `infix` notations looks like this:

```haskell
Node
    { annotations = [ ("FixitySig", "FixitySig") ]
    , children =
        [ Node
            { annotations = []
            , children = []
            , identifiers = [ OPERATOR_NAME_1 ]
            }
        , Node
            { annotations = []
            , children = []
            , identifiers = [ OPERATOR_NAME_2 ]
            }
        ]
    }
```

So, combining how AST looks like for at least these two cases, I propose the following structure. It's just an approximation

```haskell
data AstPattern
    = Anything
    | Constant Int  -- so far we have only integer constants
    | Name NameMeta Pattern
    | Parent (Set PairOfStringTags) [AstPattern]
```

After we have this data type, we can implement a matching function, and add a new type of `InspectionAnalysis`: `FindAst AstPattern`. So in the function where we traverse HIE AST, we can match on patterns.

Open questions:

1. How to name the data type? We already have `Pattern` for types...
2. What should matching function return? In case of `[0 .. length xs]` we want `SrcSpan` but for `infix` we want a list of operator names to put in a `Map`. We could return HIE `Node` type and do post-processing manually each time, since we need different info each time.

chshersh:
Resolves #107

I've implemented only types and a general matching function. To make sure that this works, an issue like #65 needs to be implemented. Probably some implementation details will change, but I decided to open a PR to show the general idea and use it in a separate PR with tests implementation.

### DIFF
--- a/src/Stan/Core/List.hs
+++ b/src/Stan/Core/List.hs
@@ -1,0 +1,30 @@
+{- |
+Copyright: (c) 2020 Kowainik
+SPDX-License-Identifier: MPL-2.0
+Maintainer: Kowainik <xrom.xkov@gmail.com>
+
+Extra functions to work with lists.
+-}
+
+module Stan.Core.List
+    ( checkWith
+    ) where
+
+{- | Checks that two lists have the same length and that a given
+binary predicate returns 'True' on each corresponding pair of
+elements.
+
+>>> checkWith (==) [] []
+True
+>>> checkWith (==) [1, 2] [1, 2]
+True
+>>> checkWith (==) [1, 2] [2, 1]
+False
+>>> checkWith (==) [1, 2] [1]
+False
+-}
+checkWith :: (a -> b -> Bool) -> [a] -> [b] -> Bool
+checkWith _ [] []         = True
+checkWith _ [] _          = False
+checkWith _ _ []          = False
+checkWith f (a:as) (b:bs) = f a b && checkWith f as bs
--- a/src/Stan/Hie/MatchAst.hs
+++ b/src/Stan/Hie/MatchAst.hs
@@ -1,0 +1,45 @@
+{- |
+Copyright: (c) 2020 Kowainik
+SPDX-License-Identifier: MPL-2.0
+Maintainer: Kowainik <xrom.xkov@gmail.com>
+
+Some 'Stan.Inspection.Inspection's require to know about AST and some
+mechanism to match parts of syntax tree to the given
+'PatternAst'. This information on AST expressions is taken from @HIE
+files@ in a more suitable view.
+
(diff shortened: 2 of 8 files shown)


src/Stan/Core/List.hs
```diff
@@ -0,0 +1,30 @@
+{- |
+Copyright: (c) 2020 Kowainik
+SPDX-License-Identifier: MPL-2.0
+Maintainer: Kowainik <xrom.xkov@gmail.com>
+
+Extra functions to work with lists.
+-}
+
+module Stan.Core.List
```
vrom911, line 9: Nice helper module!

src/Stan/Hie/MatchAst.hs
```diff
@@ -0,0 +1,45 @@
...
+SPDX-License-Identifier: MPL-2.0
+Maintainer: Kowainik <xrom.xkov@gmail.com>
+
+Some 'Stan.Inspection.Inspection's require to know about AST and some
+mechanism to match parts of syntax tree to the given
+'PatternAst'. This information on AST expressions is taken from @HIE
+files@ in a more suitable view.
+
+This module contains an implementation of the process of retrieval of this
+information from there.
```
vrom911, line 12:
```suggestion
This module contains an implementation of the process of retrieval of AST
information from @HIE@ files.
```

src/Stan/Pattern/Ast.hs
```diff
@@ -0,0 +1,33 @@
...
+    -- | AST wildcard, matches anything.
+    = PatternAstAnything
+    -- | Integer constant in code.
+    | PatternAstConstant Int  -- TODO: support constants of different types
+    -- | Name of a specific function, variable or data type.
+    | PatternAstName NameMeta
+    -- | AST node with tags for current node and children patterns
+    | PatternAstNode
+        (Set (FastString, FastString))  -- ^ Set of context info (pairs of tags)
+        [PatternAst]  -- ^ Node children
```
vrom911, line 33: Looks good :princess:

PR #117, opened 2020-05-25
[#24] Create configuration

### LINKED ISSUE #24: Create configuration
It's probably a good idea to create some configuration file, if people want to ignore some checks. I'm thinking about having `.stan.toml` file in repository with the syntax like:

```toml
ignore =
    [ { id = "HEAD" }
    , { id = "TAIL" }
    , { severity = "Warning" }
    ]
```

I'm using an array of tables of custom types. It can be written in a more verbose (but probably less noisy way):

```toml
[[ignore]]
id = "HEAD"

[[ignore]]
id = "TAIL"

[[ignore]]
severity = "Warning"
```

Basically, you can specify a list of conditions you want to ignore. In this case: all inspections with ids `HEAD` and `TAIL` and all inspections with severity `Warning` (and lower?). Let me know what do you think!

Or maybe you have more ideas about what we can put in the config? It would be also nice to be able to extend `stan` externally, so people can create their own checks (like forbid partial functions from external packages), but I feel like this deserves a separate issue and a separate discussion...

vrom911:
Resolves #24

This PR introduces `Config` data types and TOML codecs.
I will add the logic of reading config and tests separately.

I changed the types a bit in order to preserve the users' order, let me know if it still looks okay, @chshersh

### DIFF
--- a/.stan.toml
+++ b/.stan.toml
@@ -1,0 +1,19 @@
+# ignore everything in the test/ directory
+[[check]]
+type = "Ignore"
+directory = "test/"
+
+[[check]]
+type = "Include"
+
+
+# ignore specific inspection everywhere
+[[check]]
+type = "Ignore"
+inspectionId = "STAN-0002"
+
+# ignore specific inspection only in a specific file
+[[check]]
+type = "Ignore"
+inspectionId = "STAN-0001"
+file = "src/MyFile.hs"
--- a/src/Stan/Config.hs
+++ b/src/Stan/Config.hs
@@ -1,0 +1,55 @@
+{- |
+Copyright: (c) 2020 Kowainik
+SPDX-License-Identifier: MPL-2.0
+Maintainer: Kowainik <xrom.xkov@gmail.com>
+
+@stan@ configurations.
+-}
+
+module Stan.Config
+    ( -- * Data types
+      Config (..)
+    , Check (..)
+    , CheckType (..)
+    , CheckFilter (..)
+    , CheckScope (..)
+    ) where
+
+import Stan.Category (Category (..))
+import Stan.Core.Id (Id (..))
+import Stan.Core.ModuleName (ModuleName (..))
+import Stan.Inspection (Inspection (..))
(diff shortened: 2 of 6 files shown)


.stan.toml
```diff
@@ -0,0 +1,19 @@
...
+# ignore specific inspection everywhere
+[[check]]
+type = "Ignore"
+inspectionId = "STAN-0002"
+
+# ignore specific inspection only in a specific file
+[[check]]
+type = "Ignore"
+inspectionId = "STAN-0001"
+file = "src/MyFile.hs"
```
chshersh, line 19: I like, how it looks like :+1: Nice UX!

src/Stan/Config.hs
```diff
@@ -0,0 +1,55 @@
...
+    -- , configGroupBy :: !GroupBy
+    } deriving stock (Show)
+
+data CheckType
+    = Include
+    | Ignore
+    deriving stock (Show, Eq, Enum, Bounded)
+
+-- TODO: it doesn't make sense to have both Nothing, so we need a data type like 'These'
+-- data These a b = This a | That b | Both a b
```
chshersh, line 37: I think this `TODO` can be removed. I can imagine the following use-case: users comment out some parts of existing checks for some temporary experiments. In this case, checks with both `Nothing` just won't affect anything. What do you think?

src/Stan/Toml.hs
```diff
@@ -0,0 +1,98 @@
...
+checkTypeCodec = Toml.enumBounded "type"
+
+----------------------------------------------------------------------------
+-- CheckFilter
+----------------------------------------------------------------------------
+
+checkInspection :: CheckFilter -> Maybe (Id Inspection)
+checkInspection = \case
+    CheckInspection idI -> Just idI
+    _ -> Nothing
```
chshersh, line 47: Clever naming conventing with naming matching functions same as constructors :brain: I haven't thought about that, and it actually looks neat :100:

PR #131, opened 2020-05-27
[#123] CLI for specifying checks

### LINKED ISSUE #123: CLI for specifying checks
Since we have them in config, it could be a good idea to specify them in the CLI as well. The only problem: how to specify them in CLI?

I see two possible options, but I'm not sure, whether `optparse-applicative` supports them, or how to implement them.

1. **Repeat the `check` command multiple times.** The CLI will look like this:
    ```shell
    $ stan \
        check --type=Include --inspectionId=STAN-0000 \
        check --type=Ignore --observationId=bla-bla
    ```
2. **Specify arguments as a substring for the `check` option.** The CLI will look like this:
    ```shell
    $ stan \
        --check="--type=Include --inspectionId=STAN-0000" \
        --check="--type=Ignore --observationId=bla-bla"
    ```
    `optparse-applicative` has the [execParserPure](https://hackage.haskell.org/package/optparse-applicative-0.15.1.0/docs/Options-Applicative.html#v:execParserPure) function, so maybe it can be reused.

chshersh:
Resolves #123

Implementation of CLI interface for specifying checks. Implements first option from the issue.

### DIFF
--- a/src/Stan/Cli.hs
+++ b/src/Stan/Cli.hs
@@ -19,12 +19,15 @@ module Stan.Cli
 import Colourista (blue, bold, formatWith, reset, yellow)
 import Data.Version (showVersion)
 import Development.GitRev (gitCommitDate, gitHash)
-import Options.Applicative (Parser, ParserInfo (..), ParserPrefs, command, customExecParser, flag,
-                            fullDesc, help, helpLongEquals, helper, hsubparser, info, infoOption,
-                            long, metavar, prefs, progDesc, short, showDefaultWith, showHelpOnEmpty,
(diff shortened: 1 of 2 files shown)


src/Stan/Cli.hs
```diff
@@ -120,6 +125,57 @@ toggleSolutionP = flag ShowSolution HideSolution $ mconcat
     , help "Hide verbose solution information for observations"
     ]
 
+configP :: Parser Config
+configP = fmap Config $ many $ hsubparser $
+    command "check" (info checkP (progDesc "Specify list of checks"))
```
vrom911, line 130: This is cool! Using `many` for this, nice solution :+1:

PR #132, opened 2020-05-27
[#119] Read the config files

### LINKED ISSUE #119: Read the config files

vrom911:
Resolves #119

I see several issues with `Trial`:

* We actually don't always need `Last` logic in semigroup. for lists it should be (<>).
* We sometimes need to stop immediately still like when the specified file is not found).

Seems like we need to discuss more things in Trial...

### DIFF
--- a/src/Stan/Config.hs
+++ b/src/Stan/Config.hs
@@ -4,7 +4,6 @@
 {-# LANGUAGE FlexibleInstances    #-}
 {-# LANGUAGE OverloadedLabels     #-}
 {-# LANGUAGE TypeOperators        #-}
-{-# LANGUAGE TypeSynonymInstances #-}
 {-# LANGUAGE UndecidableInstances #-}
 
 {- |
--- a/src/Stan/Config.hs
+++ b/src/Stan/Config.hs
@@ -1,3 +1,12 @@
+{-# LANGUAGE ApplicativeDo        #-}
+{-# LANGUAGE DataKinds            #-}
+{-# LANGUAGE FlexibleContexts     #-}
+{-# LANGUAGE FlexibleInstances    #-}
+{-# LANGUAGE OverloadedLabels     #-}
+{-# LANGUAGE TypeOperators        #-}
+{-# LANGUAGE TypeSynonymInstances #-}
+{-# LANGUAGE UndecidableInstances #-}
+
 {- |
 Copyright: (c) 2020 Kowainik
 SPDX-License-Identifier: MPL-2.0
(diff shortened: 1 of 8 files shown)


src/Stan/Config.hs
```diff
@@ -23,10 +42,27 @@ import Stan.Observation (Observation (..))
 import Stan.Severity (Severity (..))
 
 
-data Config = Config
-    { configChecks :: ![Check]
+data ConfigP (p :: Phase Text) = ConfigP
+    { configChecks :: !(p ::- [Check])
```
chshersh, line 46: What do you think about having the list of `Trial Check` in here? In addition to existing `Trial`s on top. So we can combine history for each `Check` individually, and in the final report we can write something like:

```
Check ... comes from TOML
Check ... comes from CLI
Check ... comes from TOML
```

Since we have `Traversable` instance for `Trial`, we can convert `[Trial e a]` to `Trial e [a]` and combine all nested histories into single top-level history.

src/Stan/Config.hs
```diff
@@ -51,3 +87,13 @@ data CheckScope
...
     deriving stock (Show, Eq)
+
+defaultConfig :: PartialConfig
+defaultConfig = ConfigP
+    { configChecks = withTag "Default" $ pure []
+    }
+
+finaliseConfig :: PartialConfig -> Trial Text Config
+finaliseConfig config = do
+    configChecks <- #configChecks config
```
chshersh, line 98: Fancy features :smirk:

test/Test/Stan/Toml.hs
```diff
@@ -24,9 +25,9 @@ tomlSpec = describe "TOML configuration tests" $ do
     it "Parses test/.stan.example.toml" $
         Toml.decodeFile configCodec "test/.stan-example.toml" `shouldReturn` configExample
   where
-    configExample :: Config
-    configExample = Config
-        { configChecks =
+    configExample :: PartialConfig
+    configExample = ConfigP
+        { configChecks = ("TOML",) <$> pure
```
chshersh, line 30: Does it look batter with the `withTag` function? Or do we need `pureTagged` in `trial`? :slightly_smiling_face: 
```suggestion
        { configChecks = withTag "TOML" $ pure
```

src/Stan/Config.hs
chshersh: Turned out, I haven't thought about all details of my idea, and this might not be that helpful for users... As discussed, there are other ways to tell users in the report about checks and analysed inspections.

PR #137, opened 2020-05-28
[#135] Combine CLI options and Config options using 'trial'

### LINKED ISSUE #135: Combine CLI options and Config options using 'trial'

vrom911:
Resolves #135

I've also changed logic for default file read for TOML a bit. Now it first checks the file `.stan.toml` in the current dir, and if it is not found checks the file in the home dir.

### DIFF
--- a/src/Stan/Cli.hs
+++ b/src/Stan/Cli.hs
@@ -25,10 +25,11 @@ import Options.Applicative (Parser, ParserInfo (..), ParserPrefs, auto, command,
                             showDefaultWith, showHelpOnEmpty, showHelpOnError, strArgument,
                             strOption, subparserInline, value)
 import Options.Applicative.Help.Chunk (stringChunk)
+import Trial (fiasco, withTag)
 
 import Stan.Category (Category (..))
-import Stan.Config (Check (..), CheckFilter (..), CheckScope (..), CheckType (..), Config,
-                    ConfigP (..))
+import Stan.Config (Check (..), CheckFilter (..), CheckScope (..), CheckType (..), ConfigP (..),
+                    PartialConfig)
 import Stan.Core.Id (Id (..))
 import Stan.Core.Toggle (ToggleSolution (..))
 import Stan.Inspection (Inspection)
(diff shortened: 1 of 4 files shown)


src/Stan/Toml.hs
```diff
@@ -32,25 +34,41 @@ import qualified Toml
...
-        infoMessage $ "Configurations from " <> toText defaultTomlFile <> " will be used."
-        Toml.decodeFile configCodec file
-    else let e = fiasco "Configurations file doesn't exist" in
-        pure $ ConfigP e
+getTomlConfig = defaultCurConfigFile >>= readToml >>= \case
+    Result _ r -> pure r
+    resCur -> defaultHomeConfigFile >>= readToml >>= \ resHome ->
+        pure $ case resCur <> resHome of
+            Fiasco f     -> ConfigP $ Fiasco f
+            Result _ res -> res
```
chshersh, line 42: This is a cool idea! Looks like something where `trial` can shine. We need a function that will take a list of actions and perform them sequentially until first `Result` and also combine all histories. I'm thinking about the type signature:

```haskell
trials :: Monad m => [m a] -> (a -> m (Trial e b)) -> m (Trial e b)
```

src/Stan/Cli.hs
```diff
@@ -126,9 +127,13 @@ toggleSolutionP = flag ShowSolution HideSolution $ mconcat
...
-configP :: Parser Config
-configP = fmap ConfigP $ many $ hsubparser $
-    command "check" (info checkP (progDesc "Specify list of checks"))
+configP :: Parser PartialConfig
+configP = do
+    checks <- many $ hsubparser $
+        command "check" (info checkP (progDesc "Specify list of checks"))
+    pure $ ConfigP $ withTag "CLI" $ case checks of
+        [] -> fiasco "No CLI option specified for 'checks'"
+        xs -> pure xs
```
chshersh, line 136: Maybe we can have a function like:

```haskell
fiascoOnEmpty :: IsString tag => tag -> Parser [a] -> Parser (TaggedTrial tag [a])
```

So we can just use:

```haskell
ConfigP <$> fiascoOnEmpty "checks" (many $ hsubparser $
        command "check" (info checkP (progDesc "Specify list of checks"))
```
```

src/Stan/Toml.hs
vrom911: Nice idea! Should create an issue in `trial` :wink:

PR #138, opened 2020-05-28
[#128] [#129] Add CLI option to ignore-default and use --config-file

### LINKED ISSUE #128: CLI option to specify stan config file to use

### LINKED ISSUE #129: CLI option to ignore config file

vrom911:
Resolves #128
Resolves #129

### DIFF
--- a/src/Stan.hs
+++ b/src/Stan.hs
@@ -46,7 +46,7 @@ run = runStanCli >>= \case
 runStan :: StanArgs -> IO ()
 runStan StanArgs{..} = do
     -- config
-    tomlConfig <- getTomlConfig
+    tomlConfig <- getTomlConfig stanArgsUseDefaultConfigFile stanArgsConfigFile
     let config = finaliseConfig $ defaultConfig <> tomlConfig <> stanArgsConfig
     putTextLn $ prettyPrintTrial config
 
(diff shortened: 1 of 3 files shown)


src/Stan/Cli.hs
```diff
@@ -115,6 +119,19 @@ cabalFilePathP = many $ strOption $ mconcat
...
 
+configFileP :: Parser (Maybe FilePath)
+configFileP = optional $ strOption $ mconcat
+    [ long "config-file"
+    , metavar "FILE_PATH"
+    , help "Relative path to the .toml configurations file"
+    ]
+
+useDefaultConfigFileP :: Parser Bool
+useDefaultConfigFileP = flag True False $ mconcat
```
chshersh, line 130: `optparse-applicative` should have something like [switch](https://hackage.haskell.org/package/optparse-applicative-0.15.1.0/docs/Options-Applicative.html#v:switch) but with the reverse order of flags, like, `unswitch` or `disable` or something like that :unamused:

src/Stan.hs
```diff
@@ -46,7 +46,7 @@ run = runStanCli >>= \case
 runStan :: StanArgs -> IO ()
 runStan StanArgs{..} = do
     -- config
-    tomlConfig <- getTomlConfig
+    tomlConfig <- getTomlConfig stanArgsUseDefaultConfigFile stanArgsConfigFile
```
chshersh, line 49: I like how the main logic remains high-level!

PR #139, opened 2020-05-28
[#133] Support ENV Vars

### LINKED ISSUE #133: Support ENV Vars
I see how it could be useful for specialising default config file by default

vrom911:
Resolves #133

### DIFF
--- a/src/Stan/EnvVars.hs
+++ b/src/Stan/EnvVars.hs
@@ -28,7 +28,7 @@ getEnvVars = do
     pure EnvVars{..}
   where
     getEnvVar :: String -> (String -> Trial Text a) -> IO (TaggedTrial Text a)
-    getEnvVar var act = lookupEnv var >>= \val -> pure $ withTag "EnvVar" $ case val of
+    getEnvVar var act = lookupEnv var <&> \val -> withTag "EnvVar" $ case val of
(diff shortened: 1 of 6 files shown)


src/Stan/EnvVars.hs
```diff
@@ -0,0 +1,34 @@
...
+getEnvVars :: IO EnvVars
+getEnvVars = do
+    envVarsUseDefaultConfigFile <- getEnvVar "STAN_USE_DEFAULT_CONFIG" $ \case
+        "True"  -> pure True
+        "False" -> pure False
+        _ -> fiasco "Incorrect value is set for STAN_USE_DEFAULT_CONFIG"
+    pure EnvVars{..}
+  where
+    getEnvVar :: String -> (String -> Trial Text a) -> IO (TaggedTrial Text a)
+    getEnvVar var act = lookupEnv var >>= \val -> pure $ withTag "EnvVar" $ case val of
```
chshersh, line 31: `action >>= \a -> pure ...` can be replaced by `fmap`. Something like below should work: 

```suggestion
    getEnvVar var act = lookupEnv var <&> \val -> withTag "EnvVar" $ case val of
```

In this case, the `<&>` operator is used (reversed `<$>`). It's not that popular, so apply it only if you find it readable :slightly_smiling_face:

PR #140, opened 2020-05-28
[#120] Create the data type to accumulate and aply config info

### LINKED ISSUE #120: Create the data type to accumulate and aply config info
We will need a way to filter files/inspections/observations based on the config.

chshersh:
Resolves #120

The logic might be complicated, but I've tried to write it cleanly. Any suggestions are appreciated :slightly_smiling_face: I've decided that usage of config should be implemented in a separate issue. And I will write tests once we resolve all open questions...

**Open questions:**

1. When I work only with `[FilePath]`, I don't have info about `ModuleName`, so I can't easily apply `CheckScopeModule`...
2. Our config allows ignoring `Id Observation`. But when working only with `Inspection`, we don't have `Observation`. So our filtering should actually has 2 phases:
    1. First, we get for files only those `Inspections` we want to run.
    2. Once we get resulting observations, we should filter by observations. In this case, `HashMap FilePath (HashSet (Id Inspection))` should be changed to `HashMap FilePath ((HashSet (Id Inspection), HashSet (Id Observation)))`. Does it make sense? 
3. Some functions (e.g. filter by directory, getting all inspections by category) traverse the full list of inspections each time since we don't have fast ways to get all inspections by category or severity at the moment. I think this is okay for now, but in the future, we probably may need to optimize this somehow...

### DIFF
--- a/src/Stan/Config.hs
+++ b/src/Stan/Config.hs
@@ -188,7 +188,7 @@ applyChecks paths = foldl' useCheck filesMap
                 CheckInspection checkId -> iId == checkId
                 CheckObservation _      -> False
                 CheckSeverity sev       -> sev == inspectionSeverity
-                CheckCategory cat       -> elem cat inspectionCategory
+                CheckCategory cat       -> cat `elem` inspectionCategory
 
     applyForScope
         :: (HashSet (Id Inspection) -> HashSet (Id Inspection))
@@ -199,16 +199,10 @@ applyChecks paths = foldl' useCheck filesMap
         Nothing -> f <$> hm  -- no scope = apply for everything
         Just cScope -> case cScope of
             CheckScopeFile path -> HashMap.alter (fmap f) path hm
-            CheckScopeDirectory dir ->
-                HashMap.fromList
-                $ mapKeepMaybe
-                    (\(path, ins) -> guard (isInDir dir path) *> pure (path, f ins))
-                $ HashMap.toList hm
+            CheckScopeDirectory dir -> HashMap.mapWithKey
+                (\path -> if isInDir dir path then f else id)
+                hm
             CheckScopeModule _moduleName -> hm  -- TODO: no info about modules here :(
 
     isInDir :: FilePath -> FilePath -> Bool
     isInDir dir path = dir `isPrefixOf` path
-
--- Like 'mapMaybe' but keeps the original element if function returns Nothing
-mapKeepMaybe :: (a -> Maybe a) -> [a] -> [a]
-mapKeepMaybe f = mapMaybe (\a -> f a <|> Just a)
--- a/src/Stan/Config.hs
+++ b/src/Stan/Config.hs
@@ -32,6 +32,9 @@ module Stan.Config
 
       -- * Printing
     , configToCliCommand
+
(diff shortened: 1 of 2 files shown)


vrom911:
@chshersh Amazing work! I will answer the questions first and then review the PR more careful, as probably it could ease some things for you as well!

Sorry for not discussing this before, all the questions are raising important topics and I needed to think more about each of them :thinking: 

1. Now I see that modules indeed look redundant here. Also, that would be a nightmare to validate them as well. So I propose to remove `ChsckScopeModule` completely as it is easily replaceable with `CheckScopeFile`.
2. Interesting point! Looking into this closer I see that `Obervation`s have very different semantic in here. It actually doesn't require any filter or scope. It has 1-to-1 compliance, so it looks to me that it should be in a different fields of config rather than check, so we won't need to deal with them at this pre-stage as you mentioned.
3. Good point. Need to think more. Maybe we could somehow get the chain of filters first for each file and apply all at once? I don't know how to do this, but maybe it could be helpful to think about such structure of operation chains :chains:

chshersh:
@vrom911 Valid points! I agree with 1 and 2, this sounds reasonable. Regarding 3: I can think of a relatively simple solution for adding inspections by category or severity. We can introduce 2 top-level variables: Map from Category to inspections and Map from Severity to Inspections. So we can add inspections by Category or Severity faster. But this won't be that straightforward once we allow users to define custom inspections... Also, I still don't see a fast way to filter `HashMap FilePath` by directories...

src/Stan/Config.hs
```diff
@@ -96,3 +103,81 @@ finaliseConfig :: PartialConfig -> Trial Text Config
...
+        -> HashMap FilePath (HashSet (Id Inspection))
+    applyForScope f mScope hm = case mScope of
+        Nothing -> f <$> hm  -- no scope = apply for everything
+        Just cScope -> case cScope of
+            CheckScopeFile path -> HashMap.alter (fmap f) path hm
+            CheckScopeDirectory dir ->
+                HashMap.fromList
+                $ mapKeepMaybe
+                    (\(path, ins) -> guard (isInDir dir path) *> pure (path, f ins))
+                $ HashMap.toList hm
```
vrom911, line 175: Maybe instead of converting into list here, we can somehow use `HashMap.mapWithKey` in here?

src/Stan/Config.hs
```diff
@@ -96,3 +103,81 @@ finaliseConfig :: PartialConfig -> Trial Text Config
...
+         -- add all inspections to the existing ones
+         Nothing -> inspectionsIds <> ins
+         Just cFilter -> case cFilter of
+             CheckInspection iId -> HashSet.insert iId ins
+             CheckObservation _ -> ins
+             CheckSeverity sev ->
+                 let sevInspections = filter ((== sev) . inspectionSeverity) inspections
+                 in HashSet.fromList (map inspectionId sevInspections) <> ins
+             CheckCategory cat ->
+                 let catInspections = filter (elem cat . inspectionCategory) inspections
```
vrom911, line 146: Maybe we can make the same trick and store inspections as a `HashSet`?

Issue #148, opened 2020-05-28
Use 'Config' in the actual analysis function


chshersh:
Since we have `applyChecks` now, we can do the following:

- [x] Use `Config` in analysis
- [x] Write actual config for `stan` itself
- [x] Run `stan` on `stan` on CI

chshersh:
Once #44 and #143 is done, I think we can edit the `.stan.toml` in the following way and run `cabal run stan` on CI:

```toml
ignoreObservations =
    [ "OBS-STAN-0103-YrzpQi-27:34"
    , "OBS-STAN-0106-cpLcRq-24:15"
    , "..."
    ]

[[check]]
type = "Ignore"
directory = "target/"
```

I guess since we have the `--no-default-config` CLI flag and env var, we still will be able to test `stan` locally without many troubles.

@vrom911 What do you think?

PR #153, opened 2020-05-29
[#148] Use applyChecks in analysis functions

vrom911:
Part of #148

### DIFF
--- a/src/Stan.hs
+++ b/src/Stan.hs
@@ -28,7 +28,7 @@ import Stan.Analysis (runAnalysis)
 import Stan.Analysis.Pretty (prettyShowAnalysis)
 import Stan.Cli (InspectionArgs (..), StanArgs (..), StanCommand (..), TomlToCliArgs (..),
                  runStanCli)
-import Stan.Config (configToCliCommand, defaultConfig, finaliseConfig)
+import Stan.Config (ConfigP (..), applyChecks, configToCliCommand, defaultConfig, finaliseConfig)
 import Stan.Core.Id (Id (..))
 import Stan.EnvVars (EnvVars (..), getEnvVars)
 import Stan.Hie (readHieFiles)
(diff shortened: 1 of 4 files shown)


src/Stan.hs
```diff
@@ -55,14 +55,17 @@ runStan StanArgs{..} = do
...
-    let config = finaliseConfig $ defaultConfig <> tomlConfig <> stanArgsConfig
-    putTextLn $ prettyPrintTrial config
+    let configTrial = finaliseConfig $ defaultConfig <> tomlConfig <> stanArgsConfig
+    putTextLn $ prettyPrintTrial configTrial
 
     hieFiles <- readHieFiles stanArgsHiedir
     -- create cabal default extensions map
     cabalExtensionsMap <- createCabalExtensionsMap stanArgsCabalFilePath hieFiles
+    -- get checks for each file
+    let config = fromMaybe (error "asd") $ trialToMaybe configTrial
```
chshersh, line 65: Is it possible, that the final `configTrial` can be `Fiasco`? We have `defaultConfig` which is `pure []`. I know, that we can't prove the compiler that we won't have `Fiasco` at the end, I'm just curious, that maybe I'm missing something.

src/Stan/Analysis.hs
```diff
@@ -99,35 +102,52 @@ updateFileMap fp fi = modify' $ over fileMapL (Map.insert fp fi)
...
+            (Map.lookup hie_hs_file cabalExts)
     let fileInfoExtensions = first (ModuleParseError hie_hs_file) $
             parseSourceWithPath hie_hs_file hie_hs_src
     let fileInfoPath = hie_hs_file
     -- merge cabal and module extensions and update overall exts
     let fileInfoMergedExtensions = merge fileInfoCabalExtensions fileInfoExtensions
-    let fileInfoObservations = S.concatMap (`analysisByInspection` hieFile) inspections
+    -- get list of inspections for the file
+    let ins = catMaybes $ map lookupInspectionById $ toList insIds
+    let fileInfoObservations = S.concatMap (`analysisByInspection` hieFile) ins
```
chshersh, line 143: Oh, wow, that's quite simple and elegant implementation, nice work! :gift_heart:

src/Stan/Config.hs
```diff
@@ -146,6 +147,9 @@ configToCliCommand ConfigP{..} = "stan " <> T.intercalate " \\\n     " (map chec
         CheckScopeDirectory dir -> " --directory=" <> toText dir
         CheckScopeModule m -> " --module=" <> unModuleName m
 
+defaultChecks :: [FilePath] -> HashMap FilePath (HashSet (Id Inspection))
+defaultChecks = HashMap.fromList . map (, inspectionsIds)
+
```
chshersh, line 152: Since it's a function, not a single value, would it make sense to rename it to `mkDefaultChecks`?
Also, I guess it can be used in the `applyChecks` function now :+1:

src/Stan.hs
vrom911: It is not possible from my point of view. What do you you propose to do then? Do not have `Trial` here?

chshersh: No, `Trial` is very needed here. I'm just thinking about changing error message to `error "This is Fiasco, bro"`

PR #157, opened 2020-05-29
[#144] Write docs for applyCheck

### LINKED ISSUE #144: Write tests and docs for 'applyChecks'
And also write docs for `applyChecks`.

chshersh:
Resolves #144

### DIFF
--- a/src/Stan/Config.hs
+++ b/src/Stan/Config.hs
@@ -234,7 +234,7 @@ The 'applyConfig' function transforms the list of rules defined in the
 each module.
 
 By default, @stan@ runs all 'Inspection's for all modules in the
-Haskell project and outputs all 'Observation's it found. Using
+Haskell project and outputs all 'Observation's it finds. Using
 'Config', you can adjust the default setting using your preferences.
 
 === Algorithm
@@ -245,7 +245,7 @@ module applies each 'Check' one-by-one in order of their appearance.
 When introducing a new 'Check' in the config, you must always specify
 the 'CheckType' (either 'Include' or 'Ignore'). 'CheckFilter' and
 'CheckScope' can be omitted. If they are not written explicitly, then
-the 'Check' applies to all entries. Specifically:
+the 'Check' is applied to all the entries. Specifically:
 
 * If 'CheckFilter' is not specified, 'Check' applies to all
   inspections for the given 'CheckScope' (either 'Ignore' or 'Include'
@@ -309,7 +309,7 @@ that have a category @Partial@.
     @
 
 5. Keep 'Inspection's only with the category @Partial@ for all files
-except single one.
+except a single one.
 
     @
     # ignore all inspections
--- a/src/Stan/Config.hs
+++ b/src/Stan/Config.hs
@@ -11,7 +11,8 @@ Copyright: (c) 2020 Kowainik
 SPDX-License-Identifier: MPL-2.0
 Maintainer: Kowainik <xrom.xkov@gmail.com>
 
-@stan@ configurations.
+@stan@ runtime configuration that allows customizing the set of
+inspections to check the code against.
 -}
 
 module Stan.Config
@@ -35,6 +36,7 @@ module Stan.Config
     , configToCliCommand
 
       -- * Apply config
+      -- $applyConfig
     , applyChecks
     ) where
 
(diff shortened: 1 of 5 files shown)


src/Stan/Config.hs
```diff
@@ -207,10 +219,111 @@ applyChecks paths = foldl' useCheck (mkDefaultChecks paths)
...
     isInDir :: FilePath -> FilePath -> Bool
     isInDir dir path = dir `isPrefixOf` path
+
+{- $applyConfig
+
+The 'applyConfig' function transforms the list of rules defined in the
+'Config' (either via TOML or CLI) to get the list of 'Inspection's for
+each module.
+
+By default, @stan@ runs all 'Inspection's for all modules in the
```
vrom911, line 236: All defined in stan? Or will it be +custom? Or should we not worry about that?

src/Stan/Config.hs
```diff
@@ -207,10 +219,111 @@ applyChecks paths = foldl' useCheck (mkDefaultChecks paths)
...
+Haskell project and outputs all 'Observation's it found. Using
```
vrom911, line 237:
```suggestion
Haskell project and outputs all 'Observation's it finds. Using
```

src/Stan/Config.hs
```diff
@@ -207,10 +219,111 @@ applyChecks paths = foldl' useCheck (mkDefaultChecks paths)
...
+
+=== Algorithm
+
+The algorithm for figuring out the resulting set of 'Inspection's per
+module applies each 'Check' one-by-one in order of their appearance.
+
+When introducing a new 'Check' in the config, you must always specify
+the 'CheckType' (either 'Include' or 'Ignore'). 'CheckFilter' and
+'CheckScope' can be omitted. If they are not written explicitly, then
+the 'Check' applies to all entries. Specifically:
```
vrom911, line 248:
```suggestion
the 'Check' is applied to all the entries. Specifically:
```

src/Stan/Config.hs
```diff
@@ -207,10 +219,111 @@ applyChecks paths = foldl' useCheck (mkDefaultChecks paths)
...
+
+* If 'CheckFilter' is not specified, 'Check' applies to all
+  inspections for the given 'CheckScope' (either 'Ignore' or 'Include'
+  all 'Inspection's)
+* If 'CheckScope' is not specified, 'Check' applies given
+  'CheckFilter' to all files
+
+As a result of this approach, when both 'CheckFilter' and 'CheckScope'
+are not specified explicitly, either all 'Inspection's are 'Ignore'd,
```
vrom911, line 257: Let's rewrite it from either to the specific behaviour with each type. Otherwise it feels like it is not determined.

src/Stan/Config.hs
```diff
@@ -207,10 +219,111 @@ applyChecks paths = foldl' useCheck (mkDefaultChecks paths)
...
+or all inspections are 'Include'd back.
+
+The algorithm doesn't remove any files or inspections from the
+consideration completely. So, for example, if you ignore all
+inspections in a specific file, new inspections can be added for this
+file later by the follow up rules. If you want to ignore some files or
+directories completely, put 'Check's for them at the end of your
+configuration.
```
vrom911, line 265: What if could introduce another type of check for that? I see how it could be useful to completely remove the file from consideration (e.g. generated files)

src/Stan/Config.hs
```diff
@@ -207,10 +219,111 @@ applyChecks paths = foldl' useCheck (mkDefaultChecks paths)
...
+
+    # return back only required inspections
+    [[check]]
+    type = \"Include\"
+    category = \"Partial\"
+    file = "src/MyModule.hs"
+    @
+
+5. Keep 'Inspection's only with the category @Partial@ for all files
+except single one.
```
vrom911, line 312:
```suggestion
except a single one.
```

src/Stan/Config.hs
chshersh: In the future it will be `+custom`. But I don't think we should worry about that in the moment, we can patch the docs when we implement custom inspections...

src/Stan/Config.hs
chshersh: @vrom911 As discussed, issue #158 was created to address this problem.

src/Stan/Config.hs
chshersh: I think this section will be rewritten completely after #159...

PR #160, opened 2020-05-30
[#159] Explicit 'CheckFilter' and 'CheckScope'

### LINKED ISSUE #159: Explicit 'CheckFilter' and 'CheckScope'
As discussed with @vrom911, currently, when `CheckFilter` or `CheckScope` are not specified, the semantic of this is _all_. This can be confusing sometimes. We can introduce new constructors for both `CheckFilter` and `CheckScope` to explicitly specify the filter and scope with the **all** meaning. Specifically:

```haskell
data CheckFilter
   = ...
   | CheckAll

data CheckScope
    = ...
    | CheckScopeAll
```

And in TOML it will look like this:

```toml
[[check]]
type = "Ignore"
filter = "all"
scope = "all"
```

chshersh:
Resolves #159


src/Stan/Config.hs
```diff
@@ -242,20 +242,21 @@ Haskell project and outputs all 'Observation's it finds. Using
...
-* If 'CheckFilter' is not specified, 'Check' applies to all
-  inspections for the given 'CheckScope' (either 'Ignore' or 'Include'
-  all 'Inspection's)
-* If 'CheckScope' is not specified, 'Check' applies given
-  'CheckFilter' to all files
-
-As a result of this approach, when both 'CheckFilter' and 'CheckScope'
-are not specified explicitly, either all 'Inspection's are 'Ignore'd,
-or all inspections are 'Include'd back.
+When introducing a new 'Check' in the config, you must always three
```
vrom911, line 245:
```suggestion
When introducing a new 'Check' in the config, you must always specify three
```

src/Stan/Config.hs
```diff
@@ -242,20 +242,21 @@ Haskell project and outputs all 'Observation's it finds. Using
...
+    * 'Ignore'
+2. 'CheckFilter' — how to filter inspections
+    * 'CheckInspection': by specific 'Inspection' 'Id'
+    * 'CheckSeverity': by specific 'Severity'
+    * 'CheckCategory': by specific 'Category'
+    * 'CheckAll': applied to all 'Inspection's
+3. 'Scope' — where to apply check
+    * 'ScopeFile': only to the specific file
+    * 'ScopeDirectory': to all files in the specified directory
+    * 'ScopeAll': to all files
```
vrom911, line 259: Should we add the example of the `key-value` pair for each in here?

src/Stan/Config.hs
```diff
@@ -242,20 +242,21 @@ Haskell project and outputs all 'Observation's it finds. Using
...
 
 The algorithm doesn't remove any files or inspections from the
 consideration completely. So, for example, if you ignore all
```
vrom911, line 262: Could you please add a comment that this could be done with the `remove` fields now? :pray:

vrom911: Thank you :heartbeat:

PR #161, opened 2020-05-30
[#158] Add `remove` field in the Config

### LINKED ISSUE #158: Add `remove` field in the Config
As discussed in #157 with @chshersh , it sometimes could be useful to remove files from the `Stan`'s view at all (e.g. generated files). We can implement this feature by introducing another field in the `Config` that would contain the `CheckScope` to remove.

vrom911:
Resolves #158

I have renamed `CheckScope` -> `Scope`.

I will add `remove` tests later. For now at least property tests still pass, so it should be ok :)

### DIFF
--- a/src/Stan/Cli.hs
+++ b/src/Stan/Cli.hs
@@ -32,8 +32,8 @@ import Trial (TaggedTrial, fiasco, withTag)
 import Trial.OptparseApplicative (taggedTrialParser)
 
 import Stan.Category (Category (..))
-import Stan.Config (Check (..), CheckFilter (..), CheckScope (..), CheckType (..), ConfigP (..),
-                    PartialConfig)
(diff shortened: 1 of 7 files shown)


src/Stan/Cli.hs
```diff
@@ -166,17 +166,26 @@ toggleSolutionP = flag ShowSolution HideSolution $ mconcat
 
 configP :: Parser PartialConfig
 configP = do
-    checks <- many $ hsubparser $
-        command "check" (info checkP (progDesc "Specify list of checks"))
-    pure $ ConfigP $ withTag "CLI" $ case checks of
-        [] -> fiasco "No CLI option specified for 'checks'"
+    res <- many $ hsubparser $
+         command "check" (info (Left <$> checkP) (progDesc "Specify list of checks"))
+        <> command "remove" (info (Right <$> scopeP) (progDesc "Specify list of removed scope"))
```
chshersh, line 171: Clever solution! :open_mouth:

PR #192, opened 2020-05-31
[#106] Patterns boolean algebra

### LINKED ISSUE #106: Patterns boolean algebra
Brilliant idea by @vrom911 which I've implemented in #98, got me inspired. We could add the following constructors as well:

```haskell
PatternAnd Pattern Pattern
PatternNot Pattern
```

So we can define more flexible patterns. For example, functions `succ` and `pred` are partial for all types except `Integer`. `succ` is also not partial for `Natural`. So we could improve type patterns for them in the following way:

Current:

```haskell
succPat = (?)
predPat = (?)
```

After:

```haskell
succPat = PatternNot (integerNameMeta |-> (?)) &&& PatternNot (naturalNameMeta |-> (?))
predPat = PatternNot (integerNameMeta |-> (?))
```

### LINKED ISSUE #193: Handle negation patterns in types better
PR #192 implements boolean patterns. But they can't be used. For example, if I specify the pattern for `pred` like this:

```haskell
neg $ integer |-> (?)
```

The function of type `Integer -> Integer` matches the pattern. Because our algorithm tries to match every pattern in the list. And the `pred` function in the following code:

```haskell
foo :: Integer -> Integer
foo = pred
```

Has two types:

```haskell
1. Integer -> Integer
2. forall a . Enum a => a -> a
```

The first type doesn't match the pattern, but the second one does, that's why we still see the inspection. The second case (comparison with the type variable) happens here:

https://github.com/kowainik/stan/blob/cdd8ed9e2768796d844f3d45a35a09a5add45fa0/src/Stan/Hie/MatchType.hs#L65-L66

Something needs to be done with this to make everything works...

chshersh:
Resolves #106
Resolves #193 

The PR implements eDSL for boolean patterns, but currently I can't use them :disappointed:

### DIFF
--- a/src/Stan/Analysis/Analyser.hs
+++ b/src/Stan/Analysis/Analyser.hs
@@ -73,7 +73,9 @@ analyseNameMeta insId nameMeta patType hie@HieFile{..} =
             -- matches with the given nameMeta
             $ hieMatchNameMeta nameMeta hieId
             -- compatible with the given pattern
-            && any (hieMatchPatternType hie_types patType) typeIxs
+            && case typeIxs of
+                []    -> False
+                t : _ -> hieMatchPatternType hie_types patType t
 
         pure srcSpan
 
--- a/src/Stan/Pattern/Edsl.hs
+++ b/src/Stan/Pattern/Edsl.hs
@@ -1,0 +1,39 @@
+{- |
+Copyright: (c) 2020 Kowainik
+SPDX-License-Identifier: MPL-2.0
+Maintainer: Kowainik <xrom.xkov@gmail.com>
+
+Embedded DSL for patterns (AST and Type). Implemented using the
+/Final Taggless/ approach.
+-}
+
+module Stan.Pattern.Edsl
+    ( PatternBool (..)
+    ) where
+
+{- | Common interface for 'Bool'-like parts of patterns. Allows to
+write composable and reusable complex pattern definitions from smaller
+pieces.
+
+Laws (in terms of matching functions that return 'Bool'):
+
+* @'(?)' ≡ 'True'@
+* @'neg' '(?)' ≡ 'False'@
+* @'(?)' '|||' x ≡ x '|||' '(?)' ≡ 'True'@
+* @'(?)' '&&&' x ≡ x '&&&' '(?)' ≡ x@
+-}
+class PatternBool a where
+    -- | Anything. Matching should always return 'True'.
+    (?) :: a
+
+    -- | Negation. Inverses the argument.
+    neg :: a -> a
+
+    -- | Or-pattern. Choice.
+    (|||) :: a -> a -> a
+
+    -- | And-pattern. Both.
+    (&&&) :: a -> a -> a
+
+infixr 2 |||
+infixr 3 &&&
--- a/test/Test/Stan/Analysis/AntiPattern.hs
+++ b/test/Test/Stan/Analysis/AntiPattern.hs
@@ -17,7 +17,7 @@ analysisAntiPatternSpec analysis = describe "Anti-patterns" $ do
             "Target.AntiPattern"
             analysis
     let noObservation = noObservationAssert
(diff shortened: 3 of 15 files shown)


chshersh:
Thanks to @vrom911 idea, this works now :+1:

src/Stan/Analysis/Analyser.hs
```diff
@@ -73,7 +73,9 @@ analyseNameMeta insId nameMeta patType hie@HieFile{..} =
             -- matches with the given nameMeta
             $ hieMatchNameMeta nameMeta hieId
             -- compatible with the given pattern
-            && any (hieMatchPatternType hie_types patType) typeIxs
+            && case typeIxs of
+                []    -> False
```
vrom911, line 77: Shouldn't it be True?

test/Test/Stan/Analysis/AntiPattern.hs
```diff
@@ -17,7 +17,7 @@ analysisAntiPatternSpec analysis = describe "Anti-patterns" $ do
             "Target.AntiPattern"
             analysis
     let noObservation = noObservationAssert
-            "Target/AntiPattern.hs"
+            "target/Target/AntiPattern.hs"
```
vrom911, line 20: I guess the same fix should go to the lines above :slightly_smiling_face:

test/Test/Stan/Analysis/Partial.hs
```diff
@@ -8,14 +8,31 @@ import Stan.Analysis (Analysis)
...
+            "target/Target/Partial.hs"
+            "Target.Partial"
+            analysis
+
+    it "STAN-0010: doesn't trigger on 'succ :: Natural -> Natural'" $
+        noObservation Partial.stan0010 79
+    it "STAN-0011: doesn't trigger on 'pred :: Integer -> Integer'" $
+        noObservation Partial.stan0011 82
+
+-- TODO: configure message
```
vrom911, line 33: What do you mean here?

src/Stan/Analysis/Analyser.hs
chshersh: I tried to match behaviour of the `any` function: it returns `False` when the list is empty. Maybe it makes sense in our case as well, since if there are no types to match against, we can't match and return `True`... :thinking:

test/Test/Stan/Analysis/AntiPattern.hs
chshersh: I agree that it makes sense to pursue consistency :slightly_smiling_face: The function above doesn't need to have this prefix, since it adds `target/` inside the implementation. I guess, I'd better move this prefix inside the `noObservationAssert` function for consistency.

test/Test/Stan/Analysis/Partial.hs
chshersh: I mean, that I wanted to use `it` with the custom message, currently it duplicates. It requires some refactoring to achieve that. But I think it's worth doing it now instead of leaving a TODO, shouldn't be a big deal :+1:

vrom911: Oh, I see now, thanks!

src/Stan/Pattern/Edsl.hs
```diff
@@ -0,0 +1,39 @@
...
+{- | Common interface for 'Bool'-like parts of patterns. Allows to
+write composable and reusable complex pattern definitions from smaller
+pieces.
+
+Laws (in terms of matching functions that return 'Bool'):
+
+* @'(?)' ≡ 'True'@
+* @'neg' '(?)' ≡ 'False'@
+* @'(?)' '|||' x ≡ x '|||' '(?)' ≡ 'True'@
+* @'(?)' '&&&' x ≡ x '&&&' '(?)' ≡ x@
```
vrom911, line 23: This is cool! We need to think about how to add these tests 🤔

PR #199, opened 2020-06-02
[#198] Use type-of-html and clay for report

### LINKED ISSUE #198: Use type-of-html and clay for report

vrom911:
Resolves #198

I tried mustache but then rewrote all to `type-of-html` and `clay`. For me it seems a bit nicer, though there are few unpleasant moments:
1. I can't use `OverloadedStrings` with `type-of-html`
2. I can't write types of functions because they are too enormous and impossible to write 
:disappointed: 

@chshersh , let me know what you think about this one. Should we consider something else?


src/Stan/Report/Css.hs
```diff
@@ -0,0 +1,28 @@
...
+
+import qualified Clay as Css
+
+
+stanCss :: Css
+stanCss = do
+    Css.body ? do
+        Css.margin zero zero zero zero
+        Css.fontFamily [] [Css.sansSerif]
+        Css.color darkGrey
```
chshersh, line 15: I think, we can avoid using `qualified` import for `Clay` here. Will probably make writing CSS a bit easier

src/Stan/Report/Html.hs
```diff
@@ -0,0 +1,68 @@
+{-# OPTIONS_GHC -Wno-missing-signatures #-}
+
+{-# LANGUAGE NoOverloadedStrings #-}
```
chshersh, line 3: Heh, that's awkward... :open_mouth: :gun:

src/Stan/Report/Html.hs
```diff
@@ -0,0 +1,68 @@
...
+
+
+stanHtml an = doctype_ # html_ (stanHead # stanBody)
+  where
+    stanBody = body_ (stanHeader # stanMain an # stanFooter)
+
+stanHeader = header_ (h1_ "Stan Report")
+
+stanMain an = main_
+    ( divIdClass "general-info" ""
```
chshersh, line 35: I think, for consistency, this class can be named as others, since there's a logic for handling such words, if I understand the code correctly
```suggestion
    ( divIdClass "General Info" ""
```

src/Stan/Report/Html.hs
```diff
@@ -0,0 +1,68 @@
...
+      # divIdClassH "Analysis Info" "" ()
+      )
+    # divIdClassH "Graphs" "" (p_ "Maybe later")
+    # divIdClassH "Observations" "" (pre_ an)
+    # divIdClassH "Configurations" "" (p_ "Later")
+    # divIdClassH "Summary" "" (p_ "Later")
+    # divIdClassH "Inspections" "" (p_ "Later")
+    )
+
+divIdClass i c = div_A (A.id_ i # A.class_ c)
```
chshersh, line 47: This particular helper function looks small. Just curious, what is the type of this function will be? Is it horrible even for such small and innocent looking helpers?

src/Stan/Report/Html.hs
```diff
@@ -0,0 +1,68 @@
...
+stanHead = head_
+    ( meta_A (A.httpEquiv_ "Content-Type" # A.content_ "text/html; charset=UTF-8")
+    # meta_A (A.httpEquiv_ "X-UA-Compatible" # A.content_ "IE=Edge")
+    # nameContent "viewport" "width=device-width, initial-scale=1.0"
+    # nameContent "description" "Stan Report"
+    # nameContent "keywords" "Haskell, Static Analysis"
+    # nameContent "author" "Kowainik"
+    # title_ "Stan Report"
+
+    # style_ (render stanCss)
```
chshersh, line 65: A very clever trick of integrating `clay` and `type-of-html`?

src/Stan/Report/Html.hs
vrom911: Here I use the id without the header, so I don't put the header at all :slightly_smiling_face:

src/Stan/Report/Html.hs
vrom911: The type of the function is:

```haskell
divIdClass :: String -> String -> b -> (Div :@: ('IdA := String # 'ClassA := String)) b
```

I don't see any value in it, so I decided to omit, as it only makes the maintenance the nightmare now (as I change everything so much).. However, let me know if it would make sense to specify it explicitly.

vrom911:
@chshersh, in general, I feel it to be useful. I saw the type error, which was saying that `meta` can't be under the `html` (I forgot the `head` tag). I don't know how it would be in the long term and under heavy development. Let's see. If it would become too tedious to support, we always have already generated HTML for us which we could easily move back to `mustache` or any other templating system, I guess.

src/Stan/Report/Html.hs
chshersh: @vrom911 The type doesn't look too useful... The type signature is much bigger than the implementation, and it looks like some redundant boilerplate... I agree that it's better to omit types in this case. HTML doesn't have types either, so we're not doing worse without types, but hopefully gain some extra safety from the library and some reusability by being able to define functions.

PR #200, opened 2020-06-03
[#187] Create Observations section per file

### LINKED ISSUE #187: Create Observations section per file

vrom911:
Resolves #187


src/Stan/Observation.hs
```diff
@@ -33,7 +34,7 @@ import Stan.Core.Toggle (isHidden)
 import Stan.Hie.Debug ()
 import Stan.Inspection (Inspection (..))
 import Stan.Inspection.All (getInspectionById)
-import Stan.Report (ReportSettings (..))
+import Stan.Report.Settings (ReportSettings (..))
```
chshersh, line 37: Very nice idea to move `ReportSettings` into the separate module! This will help with incremental compilation a lot 👍

src/Stan/Observation.hs
```diff
@@ -143,6 +118,41 @@ prettyShowObservation ReportSettings{..} Observation{..} = unlines $
...
+        <> whenColour isColour reset
+      where
+        start = srcSpanStartCol observationLoc - 1
+        arrow = srcSpanEndCol observationLoc - start - 1
+
+{- | Checkes the predicate on colourfulness and returns an empty text when the
+colouroing is disabled.
+-}
+whenColour :: Bool -> Text -> Text
+whenColour isColour txt = if isColour then txt else ""
```
chshersh, line 154: Yeah, we really need to handle colouring in `colourista`... 😄

src/Stan/Report/Css.hs
```diff
@@ -9,27 +11,43 @@ module Stan.Report.Css
...
         color white
         backgroundColor lightGrey
         borderTop solid (px 8) darkGrey
+    pre ? do
+        backgroundColor black
+        color white
+        margin (2%) (10%) (2%) (10%)
+        paddingAll 2
+    ".solutions" ? do
+        margin (1%) (10%) (1%) (10%)
```
chshersh, line 38: Wow, these are fancy tricks 😏

src/Stan/Report/Html.hs
```diff
@@ -38,12 +49,55 @@ stanMain an = main_
...
-    # divIdClassH "Observations" "" (pre_ an)
+    # divIdClassH "Observations" "" (stanObservations an)
     # divIdClassH "Configurations" "" (p_ "Later")
     # divIdClassH "Summary" "" (p_ "Later")
     # divIdClassH "Inspections" "" (p_ "Later")
     )
 
+stanObservations Analysis{..} =
+    map stanPerFile
+    $ filter (not . null . fileInfoObservations)
```
chshersh, line 60: I think you can use `notNull` from `Foldable` in here
```suggestion
    $ filter (notNull . fileInfoObservations)
```

src/Stan/Report/Html.hs
```diff
@@ -38,12 +49,55 @@ stanMain an = main_
...
+    $ Map.elems analysisFileMap
+
+stanPerFile FileInfo{..} = divIdClass "" ""
+    ( h3_ (toText fileInfoPath)
+    # p_ ("Module: " # maybe mempty (unModuleName . observationModuleName) (S.safeHead fileInfoObservations))
```
chshersh, line 65: I saw a similar code in terminal pretty-printing. Do you think it's worth it to store `ModuleName` in the `FileInfo`? Or at least create a helper function like:

```haskell
extractModuleName :: Observations -> Text
```

src/Stan/Report/Html.hs
```diff
@@ -38,12 +49,55 @@ stanMain an = main_
...
+    # p_ ("Lines of Code:" <> show fileInfoLoc)
+    # divClass "observations"
+      ( strong_ "Observations" # map stanObservation (toList fileInfoObservations))
+    )
+
+stanObservation o@Observation{..} = divIdClass (unId observationId) "observation"
+    ( general
+    # pre_ (unlines $ prettyObservationSource False o)
+    # if null solutions then Nothing else Just $ divClass "solutions"
```
chshersh, line 74: This sounds like a function, that can be helpful in many places 🤔 Need to think about the name, and probably we can add it to `relude` as well 🙂 I feel like something like this should exist.

```
??? :: Bool -> a -> Maybe a
```

src/Stan/Report/Html.hs
```diff
@@ -38,12 +49,55 @@ stanMain an = main_
...
+        ( p_ "Possible solution"
+        # ul_ (map li_ solutions)
+        )
+    )
+  where
+    general = divClass "observation-general" $ ul_
+        ( li_ ("ID: " # strong_ (unId observationId))
+        # li_ ("Severity: " # show @Text (inspectionSeverity inspection))
+        # li_ ("Description: " # inspectionDescription inspection)
+        # li_ ("Inspection ID: " # a_A (A.href_ $ toText "#" <> insId) insId)
```
chshersh, line 84: How do people even write code without `OverloadedStrings` 😒

src/Stan/Report/Html.hs
```diff
@@ -38,12 +49,55 @@ stanMain an = main_
...
     # divIdClassH "Inspections" "" (p_ "Later")
     )
 
+stanObservations Analysis{..} =
+    map stanPerFile
+    $ filter (not . null . fileInfoObservations)
+    $ Map.elems analysisFileMap
+
+stanPerFile FileInfo{..} = divIdClass "" ""
+    ( h3_ (toText fileInfoPath)
```
chshersh, line 64: Will `h3_` work with `fileInfoPath` as it is? I see, that some elements can work with `Text` and some work with `String` which is a mystery for me... Why the library makes it so difficult 😞

src/Stan/Observation.hs
vrom911: Should we try the Implicit Parameters idea? :brain:

chshersh: @vrom911 I will look into the solution using the `reflection` library and will try to write a summary of this approach and a comparison between the two.

src/Stan/Report/Html.hs
vrom911: It could be just `memptyIfTrue` from `relude` if the scary type we have was a semigroup...

PR #201, opened 2020-06-04
[#197] Use relude-0.7.0.0

### LINKED ISSUE #197: Use relude-0.7.0.0

vrom911:
Resolves #197

### DIFF
--- a/src/Stan/Cabal.hs
+++ b/src/Stan/Cabal.hs
@@ -12,7 +12,7 @@ module Stan.Cabal
     , mergeParsedExtensions
     ) where
 
-import Relude.Extra.Tuple (mapToSnd)
+import Relude.Extra.Tuple (toSnd)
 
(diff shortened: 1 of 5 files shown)


src/Stan/Cabal.hs
```diff
@@ -64,7 +64,7 @@ createCabalExtensionsMap cabalPath hies = case cabalPath of
         handleCabalErr err = do
             errorMessage "Error when parsing cabal file. Stan will continue without information from .cabal file"
             pure $ Map.fromList $
-                map (mapToSnd (const $ Left $ CabalError err) . hie_hs_file) hies
+                map (toSnd (const $ Left $ CabalError err) . hie_hs_file) hies
```
chshersh, line 67: Can we use `fmapToSnd` here as well?

chshersh: Ah, I see it now 🙂

Issue #185, opened 2020-05-31
Create General Info Section


vrom911:
With three sections inside:
1. Project related General Info
2. Stan General Info
3. Analysed numbers General Info

PR #205, opened 2020-06-04
[#185] Stan Info Section

vrom911:
Part of #185


src/Stan/Info.hs
```diff
@@ -0,0 +1,94 @@
...
+
+{- |
+Copyright: (c) 2020 Kowainik
+SPDX-License-Identifier: MPL-2.0
+Maintainer: Kowainik <xrom.xkov@gmail.com>
+
+@stan@ build information.
+-}
+
+module Stan.Info
```
chshersh, line 11: This is a great module! :top:

PR #206, opened 2020-06-04
[#100] Inspection: missing `infix` for operator

### LINKED ISSUE #100: Inspection: missing `infix` for operator
Analysis idea: collect two sets:

* All operator definitions
* All `infix` definitions 

Then, compare at the end, which operators are not in the `infix` set, and report observation for them.

chshersh:
Resolves #100

A new inspection looks like this:

Screenshot from 2020-06-04 22-53-52

The heaviest module is `Stan.Analysis.Analyser`. It contains the main logic for processing the AST and collecting all operators.

What else is done:

* New `Stan.Inspection.Style` module for the `Style :: Severity` inspections
* New constructor to `PatternAst` to match on any number of children nodes
* New AST patterns to handle fixities

Open questions and what else needs to be done:

1. More tests when the inspection doesn't trigger?
2. Currently, doesn't handle operators as a part of typeclasses (no warning on `(?)` from `PatternBool` eDSL).
3. Also warns on operators inside `where`.

### DIFF
--- a/src/Stan/Analysis/Analyser.hs
+++ b/src/Stan/Analysis/Analyser.hs
@@ -29,6 +29,7 @@ import Stan.Observation (Observations, mkObservation)
 import Stan.Pattern.Ast (PatternAst, fixity, typeSig)
 import Stan.Pattern.Type (PatternType)
 
+import qualified Data.HashMap.Strict as HM
 import qualified Data.Map.Strict as Map
 import qualified Slist as S
 
@@ -115,7 +116,7 @@ analyseInfix
 analyseInfix insId hie =
     let opDecls = analyseAstWith (matchInfix <> matchOperator) hie
         (fixities, topOperators) = partitionDecls opDecls
-        operatorsWithoutFixity = Map.difference topOperators fixities
+        operatorsWithoutFixity = HM.difference topOperators fixities
     in mkObservation insId hie <$> slist (toList operatorsWithoutFixity)
   where
     -- returns list of operator names defined in a single fixity declaration:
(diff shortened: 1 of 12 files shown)


target/Target/Style.hs
```diff
@@ -0,0 +1,7 @@
+{-# OPTIONS_GHC -fno-warn-missing-export-lists #-}
+
+module Target.Style where
+
+
+(???) :: Int -> Int -> Int
+(???) = (+)
```
vrom911, line 7: Should we also check that it is not triggered on the existing fixity?

src/Stan/Hie/MatchAst.hs
```diff
@@ -57,7 +57,9 @@ hieMatchPatternAst hie@HieFile{..} node@Node{..} = \case
         any (matchNameAndType nameMeta patType)
         $ Map.assocs
         $ nodeIdentifiers nodeInfo
-    PatternAstNode tags patChildren ->
+    PatternAstNode tags ->
+        tags `Set.isSubsetOf` nodeAnnotations nodeInfo
+    PatternAstNodeExact tags patChildren ->
            tags `Set.isSubsetOf` nodeAnnotations nodeInfo
```
chshersh, line 63: Maybe move into the separate function

src/Stan/Analysis/Analyser.hs
```diff
@@ -86,16 +91,114 @@ analyseAst
...
+    | Operator Text RealSrcSpan
+
+{- | Partition a foldable of operator declarations into two maps:
+
+1. Fixity declarations (mapped to @()@).
+2. Top-level operator names (mapped to their source positions.
+
+'Map' is used to be able to use the nice @merge@ function.
+-}
+partitionDecls :: Foldable f => f OperatorDecl -> (Map Text (), Map Text RealSrcSpan)
```
chshersh, line 176: Use `HashMap` instead of `Map`

PR #207, opened 2020-06-05
[#185] Create General Info Section

### LINKED ISSUE #185: Create General Info Section

vrom911:
Resolves #185


src/Stan.hs
```diff
@@ -73,13 +75,19 @@ runStan StanArgs{..} = do
         putTextLn res
 
         when stanArgsReport $ do
+            -- Project Info
+            piName <- takeFileName <$> getCurrentDirectory
+            piCabalFiles <- usedCabalFiles stanArgsCabalFilePath
+            let piHieDir = stanArgsHiedir
+            let piFileNumber = length hieFiles
```
chshersh, line 82: Hmm, I just thought: would it make sense for the `readHieFiles` function to return `Slist` instead of ordinary list? I think we already call the `length` function several times, and with this approach we can have fewer observations in the `stan` itself 🙂 I guess this can be done as a separate PR, just wanted to know what do you think about this.

vrom911: Sounds like a good idea! :+1: :open_mouth:

Issue #99, opened 2020-05-03
Inspections for known anti-patterns


chshersh:
- [x] `foldl`
- [x] `Data.ByteString.Char8.pack`
- [x] Slow `O(n)` `HashMap.size`, `length` (both strict and lazy, if there's difference)
- [x] Slow `O(n)` `HashSet.size`, `length` (both strict and lazy, if there's difference)
- [x] Slow `O(n)` `Text.length`
- [x] Slow `O(n^2)` `nub`
- [x] `for_ [1..10000000]`
- [ ] `(</>)` for urls

PR #212, opened 2020-06-05
[#99] Anti-pattern: Data.ByteString.Char8.pack

chshersh:
Part of #99


src/Stan/NameMeta.hs
```diff
@@ -60,7 +60,7 @@ compareNames NameMeta{..} name =
         isExternalName name
         && occName    == nameMetaName
         && moduleName == nameMetaModuleName
-        && package    == nameMetaPackage
+        && nameMetaPackage `T.isPrefixOf` package
```
vrom911, line 63: Heh, need to investigate this one :mag:

Issue #87, opened 2020-05-01
Handle multiple definitions for Inspection


chshersh:
Some functions can come from multiple places. For example, `length` can come from `GHC.OldList` as a monomorphic function, or as polymorphic function, specialised to lists from `Data.Foldable`. I guess in such cases we want to somehow support multiple definition points for a single function. A naive solution would be to store `NonEmpty InspectionAnalysis` instead of `InspectionAnalysis` inside `Inspection` but maybe there's a better solution.

- [ ] length
- [ ] fromList

chshersh:
I think that recent @vrom911 idea that I used in #98 probably can be applied here as well in some form. We can add a new constructor `InspectionOr InspectionAnalysis InspectionAnalysis` to the `InspectionAnalysis` type. Though, I'm a bit worried that this solution is too permissive. I don't think it makes sense to define `InspectionOr Infix (FindName ..)`. What we want is some list of names and corresponding type patterns.

Currently, we have: `FindName NameMeta Pattern`. I wonder if can somehow have `Or` only for pair of `NameMeta` and `Pattern` :thinking:

chshersh:
This issue is automatically done once we have #237 implemented, since the `PatternAst` inside `FindAst` already has boolean algebra implemented inside. The only thing that left to be done in this issue: support multiple functions for different data types, specifically:

- [ ] `fromList` for `NonEmpty` can come either from `IsList` or `Data.List.NonEmpty`
- [ ] Support both `size` and `length` for the `HashSet.size` inspection
- [ ] Support both `size` and `length` for the `HashMap.size` inspection
- [ ] Add tests

Separate consideration: all functions that work with infinite lists, can also work with infinite `NonEmpty`. Should we patch the rules for them as well?

PR #213, opened 2020-06-05
[#99] Anti-pattern: slow HashMap.size

chshersh:
Part of #99

### DIFF
--- a/src/Stan/Inspection/AntiPattern.hs
+++ b/src/Stan/Inspection/AntiPattern.hs
@@ -19,6 +19,8 @@ module Stan.Inspection.AntiPattern
     , stan0202
       -- *** Anti-pattern 'Data.ByteString.Char8.pack'
     , stan0203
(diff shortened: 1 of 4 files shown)


src/Stan/Inspection/AntiPattern.hs
```diff
@@ -98,10 +109,20 @@ stan0203 = mkAntiPatternInspection (Id "STAN-0203") "Data.ByteString.Char8.pack"
...
         }
 
-lenPatAst :: PatternAst
-lenPatAst = range
-    (PatternAstConstant 0)
-    (app
-        (PatternAstName (mkBaseFoldableMeta "length") (?))
-        (?)
-    )
+-- TODO: also warn on 'length'
```
chshersh, line 112: There's already a separate issue to support such a way of defining inspections: https://github.com/kowainik/stan/issues/87

PR #214, opened 2020-06-05
[#195] Implement `cli-to-toml` command

### LINKED ISSUE #195: Implement `cli-to-toml` command

chshersh:
Resolves #195

That was quite straightforward, thanks to `tomland` and previous work by @vrom911 :slightly_smiling_face:

### DIFF
--- a/src/Stan.hs
+++ b/src/Stan.hs
@@ -17,7 +17,7 @@ module Stan
 
 import Colourista (errorMessage, formatWith, infoMessage, italic)
 import HieTypes (HieFile (..))
-import System.Directory (getCurrentDirectory)
+import System.Directory (doesFileExist, getCurrentDirectory)
 import System.Environment (getArgs)
 import System.FilePath (takeFileName)
 import Trial (pattern FiascoL, pattern ResultL, Trial (..), prettyPrintTaggedTrial,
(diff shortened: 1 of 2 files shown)


src/Stan.hs
```diff
@@ -113,3 +114,12 @@ runTomlToCli TomlToCliArgs{..} = do
         fiasco -> do
             errorMessage "Could not get Configurations:"
             putTextLn $ prettyPrintTrial fiasco
+
+runCliToToml :: CliToTomlArgs -> IO ()
+runCliToToml CliToTomlArgs{..} = do
+    let toml = Toml.encode configCodec cliToTomlArgsConfig
+    case cliToTomlArgsFilePath of
+        Nothing   -> putTextLn toml
```
vrom911, line 122: Let's add an info message, something like, copy-paste this into `.stan.toml` and Stan will pick up this file on the next run. Or maybe say about any file and remind about --config-file option? :thinking:

src/Stan.hs
```diff
@@ -113,3 +114,12 @@ runTomlToCli TomlToCliArgs{..} = do
...
+        Just path -> do
+            infoMessage $ "TOML configuration is written to file: " <> toText path
```
vrom911, line 124: Should we first check if the file exist?

src/Stan.hs
```diff
@@ -113,3 +114,12 @@ runTomlToCli TomlToCliArgs{..} = do
...
+            writeFileText path toml
```
vrom911, line 125: I guess these two lines can be switched over in order :slightly_smiling_face:

PR #216, opened 2020-06-06
Add short description to each section

chshersh:


src/Stan/Report/Html.hs
```diff
@@ -16,6 +16,7 @@ module Stan.Report.Html
 
 import Clay (compact, renderWith)
 import Data.Char (toLower)
+import Html
```
chshersh, line 19: Temporary import to make conflict resolution easier. But feel free to make import for the `Html` module implicit :slightly_smiling_face: I feel like these libraries (`type-of-html` and `clay`) require enormous imports because of how they were designed...

src/Stan/Report/Css.hs
```diff
@@ -76,6 +77,10 @@ stanCss = do
     ".severityWarning" ? backgroundColor yellow
     ".severityError" ? backgroundColor red
 
+    blockquote ? do
+        paddingLeft (2%)
+        borderLeft solid (px 4) darkGrey
```
vrom911, line 82: Let's use `box-shadow` to have double borderline with grey and light grey at once :nail_care:

PR #220, opened 2020-06-06
[#182] Calculate project health

### LINKED ISSUE #182: Calculate project health
How can we count project health? I imagine some formula depending on the total number of inspections, the total number of LOC and the number of found observations.

chshersh:
Resolves #182

### DIFF
--- a/src/Stan/Analysis/Pretty.hs
+++ b/src/Stan/Analysis/Pretty.hs
@@ -11,10 +11,14 @@ module Stan.Analysis.Pretty
 
       -- * Numbers
     , AnalysisNumbers (..)
+    , ProjectHealth (..)
     , analysisToNumbers
+    , prettyHealth
+    , toProjectHealth
     ) where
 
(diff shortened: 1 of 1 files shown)


src/Stan/Analysis/Pretty.hs
```diff
@@ -62,8 +67,49 @@ analysisToNumbers Analysis{..} = AnalysisNumbers
...
+    | MediumHealth
+    | Healthy
+
+-- | Calculate 'ProjectHealth'.
+toProjectHealth :: Double -> ProjectHealth
+toProjectHealth health
+    | health >= 100 = Healthy
+    | health >= 80  = MediumHealth
+    | health >= 40  = LowHealth
+    | otherwise     = Unhealthy
```
chshersh, line 112: Currently, this function is not used. I've implemented it to display health better later in HTML. Let me know, if it can be used somehow in CLI reports as well :slightly_smiling_face:

PR #228, opened 2020-06-07
[#190] Create Summary Section

### LINKED ISSUE #190: Create Summary Section

chshersh:
Resolves #190

### DIFF
--- a/src/Stan/Report/Html.hs
+++ b/src/Stan/Report/Html.hs
@@ -18,10 +18,12 @@ import Relude.Extra.Enum (universe)
 
 import Clay (compact, renderWith)
 import Data.Char (toLower)
-import Html
+import Html hiding (Summary)
 
 import Stan.Analysis (Analysis (..))
-import Stan.Analysis.Pretty (AnalysisNumbers (..), analysisToNumbers)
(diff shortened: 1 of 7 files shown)


src/Stan/Report/Html.hs
```diff
@@ -133,6 +139,38 @@ stanAnalysis (analysisToNumbers -> AnalysisNumbers{..}) =
...
+        # summary
+        )
+    )
+  where
+    showProjectHealth :: ProjectHealth -> Text
+    showProjectHealth = toText . \case
+        Unhealthy    -> "is unhealthy"
+        LowHealth    -> "has low health"
+        MediumHealth -> "has medium health"
+        Healthy      -> "is healthy"
```
vrom911, line 156: Need to think on wording to make it more helpful and less blaming 😃 Also we can add some advising

src/Stan/Report/Html.hs
```diff
@@ -133,6 +139,38 @@ stanAnalysis (analysisToNumbers -> AnalysisNumbers{..}) =
...
+
+    summary = case createSummary analysis of
+        Nothing -> Left $ li_ "Congratulations! Your project has zero vulnerabilities!"
```
vrom911, line 159: Maybe "well done" or "good job" fits here better? It is their achievement, and we can highlight that, what do you think?

PR #233, opened 2020-06-08
[#171] Inspection: strict data type fields

### LINKED ISSUE #171: Inspection: strict data type fields

chshersh:
Resolves #171

This was tough :sweat: 

Implemented inspection for finding lazy fields and producing a warning on them. Current limitation: ignores `data` with a single field. HIE AST doesn't track difference between `newtype` and `data`, so some workarounds involved.

TODO:

* Fix warnings for our own code (make all fields strict in all data types). The inspection has already proven itself handy!

### DIFF
--- a/src/Stan/FileInfo.hs
+++ b/src/Stan/FileInfo.hs
@@ -53,5 +53,5 @@ isExtensionDisabled :: Extension -> ExtensionsResult -> Bool
 isExtensionDisabled ext = \case
     Left _ -> True  -- no info about extensions, consider it disabled
     Right Extensions{..} ->
-           Set.member (Off ext) extensionsAll
-        || Set.notMember (On ext) extensionsAll
+           Set.notMember (On ext) extensionsAll
+        || Set.member (Off ext) extensionsAll
--- a/src/Stan/FileInfo.hs
+++ b/src/Stan/FileInfo.hs
@@ -11,12 +11,17 @@ module Stan.FileInfo
     , FileInfo (..)
 
     , extensionsToText
+    , isExtensionDisabled
     ) where
 
-import Extensions (ExtensionsError, ExtensionsResult, ParsedExtensions (..), showOnOffExtension)
+import Extensions (Extensions (..), ExtensionsError, ExtensionsResult, OnOffExtension (..),
+                   ParsedExtensions (..), showOnOffExtension)
+import GHC.LanguageExtensions.Type (Extension)
 
 import Stan.Core.ModuleName (ModuleName)
 import Stan.Observation (Observations)
+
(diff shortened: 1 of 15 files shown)


src/Stan/FileInfo.hs
```diff
@@ -41,3 +46,12 @@ extensionsToText = \case
...
             Nothing -> exts
+
+{- | Check whether the given extension is disabled
+-}
+isExtensionDisabled :: Extension -> ExtensionsResult -> Bool
+isExtensionDisabled ext = \case
+    Left _ -> True  -- no info about extensions, consider it disabled
+    Right Extensions{..} ->
+           Set.member (Off ext) extensionsAll
+        || Set.notMember (On ext) extensionsAll
```
vrom911, line 57: I guess the second case is more likely to be there, so I think it makes sense to check it first

src/Stan/Pattern/Ast.hs
```diff
@@ -81,11 +84,69 @@ infixr 7 ***, +++, ???
...
+-}
+lazyField :: PatternAst
+lazyField = lazyRecordField ||| type_
+
+{- | Pattern for any occurrence of a plain type. Covers the following
+cases:
+
+* Simple type: Int, Bool, a
+* Higher-kinded type: Maybe Int, Either String a
+* Type in parenthesis: (Int)
```
vrom911, line 120: I s this really a separate case? :thinking:

src/Stan/Pattern/Ast.hs
```diff
@@ -81,11 +84,69 @@ infixr 7 ***, +++, ???
...
+* Tuples: (Int, Bool)
+* List type: [Int]
```
vrom911, line 122: Are there any other special types?
Maybe functions?

test/Test/Stan/Number.hs
```diff
@@ -17,4 +17,4 @@ linesOfCodeSpec hieFile = describe "LoC tests" $
 modulesNumSpec :: Int -> Spec
 modulesNumSpec num = describe "Modules number tests" $
     it "should count correct number of modules" $
-        num `shouldBe` 56
+        num `shouldSatisfy` (> 56)
```
vrom911, line 20: :unamused:

src/Stan/Pattern/Ast.hs
chshersh: Yes, it is, and I was surprised by this as well :unamused:

src/Stan/Pattern/Ast.hs
chshersh: Good catch! The function is indeed one more builtin I haven't considered

PR #245, opened 2020-06-11
[#99] Add length for Text inspection

vrom911:


.stan.toml
```diff
@@ -1,2 +1,7 @@
 [[remove]]
 directory = "target/"
+
+[[check]]
+type = "Exclude"
+directory = "test/"
+id = "STAN-0208"
```
chshersh, line 7: Using more Stan features in the project itself! 😍

PR #248, opened 2020-06-11
[#240] Bypass Cabal hack on MacOS

### LINKED ISSUE #240: Bypass Cabal hack on MacOS
:fire: :fire: :fire: 

Currently, some of the inspections won't work on MacOS due to the logic of Cabal `dist-newstyle` packages naming convention on MacOS. Turns out, they are removing all vowels from the package name on MacOS only. 

This is crucial for us as we need to know what package the function is coming from to check the inspection.

P.S. I was so lucky to remember this comment I saw once on Reddit https://www.reddit.com/r/haskell/comments/gy0lsv/making_the_most_of_cabal/ft95rqs/

Without this I won't be able to figure out the issue on MacOS, like _never_ :skull:

vrom911:
Resolves #240

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,10 +10,11 @@ on:
 
 jobs:
   build:
-    name: ghc ${{ matrix.ghc }}
-    runs-on: ubuntu-16.04
+    name: ${{ matrix.os }} / ghc ${{ matrix.ghc }}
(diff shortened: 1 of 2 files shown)


vrom911:
Regarding the latest commit, see this issue for GHC-8.10 on CI for MacOS: https://github.com/actions/setup-haskell/issues/21

PR #257, opened 2020-06-12
[#255] Forbid unsafe functions that could lead to errors

### LINKED ISSUE #255: Forbid unsafe functions that could lead to errors
- [ ] `undefined`
- [ ] `unsafeCoerce`
- [ ] `unsafePerformIO`
- [ ] `unsafeInterleaveIO`
- [ ] `unsafeDupablePerformIO`

Should they be separate inspections? Like 3 inspections with different solutions:

1. At least use `error` instead of `undefined`
2. Use `coerce` instead of `unsafeCoerce`
3. Using `unsafeIO` functions breaks referential transparency.

chshersh:
Resolves #255

### DIFF
--- a/.stan.toml
+++ b/.stan.toml
@@ -5,3 +5,8 @@ directory = "target/"
 type = "Exclude"
 directory = "test/"
 id = "STAN-0208"
+
+[[check]]
+type = "Exclude"
+file = "src/Stan/Cabal.hs"
+id = "STAN-0212"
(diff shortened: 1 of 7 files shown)


.stan.toml
```diff
@@ -4,4 +4,9 @@ directory = "target/"
 [[check]]
 type = "Exclude"
 directory = "test/"
-id = "STAN-0208"
\ No newline at end of file
+id = "STAN-0208"
+
+[[check]]
+type = "Exclude"
+file = "src/Stan/Cabal.hs"
```
vrom911, line 11: Could you add the comment in this module explaining why it is important to leave `unsafeInterleave` in there?

Issue #130, opened 2020-05-27
Custom formatting for CLI --help


chshersh:
Implement custom formatting of the `--help` output.

- [x] Pretty header
- [x] **Optional** Pretty footer
- [x] Use colors, italic and bold emphasis
- [x] Group options
- [x] Better `--help` on errors

Also, think about other possible improvements for implementing custom `--help` output.

PR #258, opened 2020-06-12
[#130] Some CLI improvements

vrom911:
Question: should we use colours and any other formatting? I tried bold on commands now, could you please check it out and leave your feedback on how it looks.

On a bigger note, unfortunately, the `optparse-applicative` library's functional is not very flexible in prettifying meaning. Very limited amount of pieces can be configured and absolutely no possibility to change the Usage section

### DIFF
--- a/src/Stan/Cli.hs
+++ b/src/Stan/Cli.hs
@@ -94,7 +94,7 @@ stanParserPrefs = prefs $ mconcat
 
 stanCliParser :: ParserInfo StanCommand
 stanCliParser = modifyHeader $ info (helper <*> versionP <*> stan) $
-    fullDesc <> progDesc "Haskell Static Analyser"
+    fullDesc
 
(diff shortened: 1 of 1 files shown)


src/Stan/Cli.hs
```diff
@@ -221,6 +235,15 @@ configP = do
             , configIgnored = whenEmpty ignored "ignore"
             }
   where
+    cmd :: String -> String -> (a -> ConfigCommand) -> Parser a -> Parser ConfigCommand
+    cmd name h cc p = hsubparser
+        ( command name
+            (info (cc <$> p) (progDesc h))
+        <> commandVar (map toUpper name) -- TODO: should we: <> " !{--" <> name <> "-options}")
```
chshersh, line 242: I think it's okay to keep as it is :+1:

PR #259, opened 2020-06-13
[#242] Support GHC-8.10

### LINKED ISSUE #242: Support GHC-8.10
We don't use much of GHC API, so it shouldn't be that hard. But if some API changes, I propose to handle this in the following way:

1. Create `Stan.Hie.Compat` with `CPP` around relevant functions to work with HIE. Ideally, the implementations of functions from different GHC versions are allowed to be different, but the types should be the same whenever it's possible to achieve it, so we can use them in other modules and have CPP only in `Stan.Hie.Compat`.
2. Optionally, if some other parts of GHC change, create the `Stan.Ghc` (or `Stan.Ghc.Compat`) module and do the same as for `Stan.Hie.Compat`.

I hope that none of this will be required since we don't use unstable parts of the GHC API. But it's a plan to keep in mind.

chshersh:
Resolves #242

That was unexpectedly easy...

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,13 +16,18 @@ jobs:
       matrix:
         os: [ubuntu-latest, macOS-latest]
         cabal: ["3.0"]
-        ghc: ["8.8.3"]
+        ghc:
+          - "8.8.3"
+          - "8.10.1"
+        exclude:
+          - os: macOS-latest
+            ghc: 8.8.3
 
     steps:
     - uses: actions/checkout@v2
       if: github.event.action == 'opened' || github.event.action == 'synchronize' || github.event.ref == 'refs/heads/master'
 
-    - uses: actions/setup-haskell@v1.1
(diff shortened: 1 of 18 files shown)


vrom911:
Did you decide not to stick to the plan? Why so? In that case, you better to write down reasons not to do that.

chshersh:
@vrom911 In the #242 issue I've described the plan to support compatibility across multiple GHC versions because of API of different GHCs _may_ change in various ways (module renaming, types renaming, changing of types, change of behaviour, etc.). In that case, it will be more convenient to put definitions of types and functions with API changes under CPP in separate modules `Stan.Ghc.Compat` and `Stan.Hie.Compat`. But, apparently, nothing changed between GHC-8.8.3 and GHC-8.10.1 in modules that we use. So I don't know what to put in those modules... I'm afraid, that in the current situation, introducing new modules will bring more overhead without solving the real issue. Maybe API changes of future GHC-8.12 will be more dramatic, and we will have to introduce compatibility modules and put all imports in those modules.

PR #260, opened 2020-06-13
[#256] Clean up CLI output

### LINKED ISSUE #256: Clean up CLI output
Currently, env vars `Trial` is outputted without explanations, and we need to patch few other things as well. including `Config` data type. Maybe we can make it in the report style as we already have all of the functions.

vrom911:
Resolves #256

### DIFF
--- a/src/Stan/Config/Pretty.hs
+++ b/src/Stan/Config/Pretty.hs
@@ -10,9 +10,13 @@ module Stan.Config.Pretty
     ( ConfigAction (..)
     , prettyConfigAction
     , configActionClass
+    , configActionColour
+    , prettyConfigCli
 
     , configToTriples
     ) where
+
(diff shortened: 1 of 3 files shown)


src/Stan/Config/Pretty.hs
```diff
@@ -70,3 +81,12 @@ prettyScope = \case
...
     ScopeAll -> "All files"
+
+prettyConfigCli :: Config -> Text
+prettyConfigCli = unlines . concatMap action . configToTriples
+  where
+    action :: (ConfigAction, Text, Text) -> [Text]
+    action (act, check, scope) =
+          formatWith [configActionColour act, bold] (prettyConfigAction act)
+        :  [ "    " <> check | check /= ""]
+        ++ [ "    " <> scope | scope /= ""]
```
chshersh, line 92: Wow, this is very nice code!

PR #262, opened 2020-06-13
[#261] Generate md with all inspections

### LINKED ISSUE #261: Generate md with all inspections
We need a place to store all our inspections with description and all that stuff.
For starters, I think we can host in on our Wiki page. But writing it manually is a bit of overhead, so I propose to write a command in CLI that will do that for us.

vrom911:
Resolves #261

### DIFF
--- a/src/Stan/Cli.hs
+++ b/src/Stan/Cli.hs
@@ -24,10 +24,10 @@ import Colourista.Short (b)
 import Data.Char (toUpper)
 import Options.Applicative (CommandFields, Mod, Parser, ParserInfo (..), ParserPrefs, auto, columns,
                             command, commandGroup, customExecParser, flag, flag', fullDesc, help,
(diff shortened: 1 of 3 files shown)


src/Stan/Cli.hs
```diff
@@ -156,6 +161,15 @@ stanCliToTomlP = hsubparser $ commandGroup "TOML Configurations"
...
         pure $ StanCliToToml CliToTomlArgs{..}
 
+stanInspectionsToMd :: Parser StanCommand
+stanInspectionsToMd = hsubparser
+    $  command "inspections-to-md"
+        (info (pure StanInspectionsToMd)
+             (progDesc "Create md with all inspections info")
+        )
+    <> help "Create md with all inspections info"
+    <> hidden <> internal
```
chshersh, line 171: :male_detective:

PR #269, opened 2020-06-14
[#173] Inspection: replace multiple comparison operations with operators

### LINKED ISSUE #173: Inspection: replace multiple comparison operations with operators
When starting to learn Haskell, you can write code like this:

```haskell
foo x k t
    | x < k = ...
    | x > k = ...
    | otherwise = ...
```

It's more idiomatic (and efficient for composite data types) to use `compare`.

chshersh:
Resolves #173

### DIFF
--- a/src/Stan/Analysis/Analyser.hs
+++ b/src/Stan/Analysis/Analyser.hs
@@ -18,17 +18,21 @@ import GHC.LanguageExtensions.Type (Extension (Strict, StrictData))
 import Slist (Slist, slist)
 
 import Stan.Core.Id (Id)
+import Stan.Core.List (nonRepeatingPairs)
 import Stan.FileInfo (isExtensionDisabled)
 import Stan.Ghc.Compat (RealSrcSpan, isSymOcc, nameOccName, occNameString)
+import Stan.Hie (eqAst)
 import Stan.Hie.Compat (HieAST (..), HieASTs (..), HieFile (..), Identifier, NodeInfo (..),
                         TypeIndex)
 import Stan.Hie.MatchAst (hieMatchPatternAst)
 import Stan.Inspection (Inspection (..), InspectionAnalysis (..))
+import Stan.NameMeta (NameMeta, ghcPrimNameFrom)
 import Stan.Observation (Observations, mkObservation)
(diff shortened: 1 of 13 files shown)


src/Stan/Pattern/Ast.hs
```diff
@@ -193,3 +220,25 @@ tuple =
...
+guardBranch = PatternAstNode $ one ("BodyStmt", "StmtLR")
+
+{- | Pattern for the right-hand-side. Usually an equality sign.
+
+@
+   foo = baz
+@
+-}
+rhs :: PatternAst
+rhs = PatternAstNode $ one ("GRHS", "GRHS")
```
vrom911, line 240: Oh, that what it means! I used the same and called it `patternMatchingArrow` assuming that it would be used only there. I can reuse it!

test/Test/Stan/Analysis/AntiPattern.hs
```diff
@@ -130,3 +131,39 @@ unsafeFunctionsSpec analysis = describe "STAN-0212: Unsafe functions" $ do
...
+        noObservation 26
+        noObservation 27
+    it "No warning on: >= with different constants" $ do
+        noObservation 30
+        noObservation 31
+        noObservation 32
+        noObservation 33
+        noObservation 34
+        noObservation 35
+    xit "Handles functions with pattern-matching" $
```
vrom911, line 168: :+1: can figure out this one later! This is already huge and amazing :tangerine:

src/Stan/Analysis/Analyser.hs
```diff
@@ -78,6 +84,72 @@ analyseBigTuples insId hie =
...
+        -> Maybe (HieAST TypeIndex, HieAST TypeIndex)
+    extractComparisonGuard node = do
+        -- guard starts with GRHS annotation
+        guard $ hieMatchPatternAst hie node rhs
+        -- guard predicate is a first son
+        stmt:_ <- Just $ nodeChildren node
+        -- check if it's a guard
+        guard $ hieMatchPatternAst hie stmt guardBranch
+        -- check if it's an operator
+        guard $ hieMatchPatternAst hie stmt $ opApp (?) opsPat (?)
```
vrom911, line 122: Nice guarding of guards :guard:

PR #272, opened 2020-06-15
[#35] Add --non-verbose option in CLI

### LINKED ISSUE #35: Add --non-verbose option in CLI
For very short one-line output per observation:

```
$ stan --non-verbose
 * [1-STAN-001] WARNING <src/Example.hs(20:4-20:8)> Partial: head
 * [2-STAN-001] ERROR   <src/Example.hs(20:4-20:8)> Partial: head
```
^ the output format should be discussed better :slightly_smiling_face:

vrom911:
Resolves #35


src/Stan/Report/Settings.hs
```diff
@@ -8,13 +8,40 @@ Report settings types.
...
+    | NonVerbose
+    deriving stock (Show)
+
+isVerbose :: Verbosity -> Bool
+isVerbose = \case
+     Verbose -> True
+     NonVerbose -> False
+
+-- | Boolean for showing/hiding solution information of observations.
+data ToggleSolution
```
vrom911, line 39: @chshersh , do you think that I should use verbosity instead of toggle here?

chshersh: @vrom911 No, I don't think so. I think it's okay to have different `Bool`-like enums to make code more readable.

PR #278, opened 2020-06-18
[#275] Inspection: invalid file paths on Windows

### LINKED ISSUE #275: Inspection: invalid file paths on Windows
When paths are hardcoded like this `"foo/bar/baz"` they won't work on Windows. We can provide an inspection to check this.

The only question: where to search for such strings? I think the easiest solution in the first version is to check string literals against both arguments of `</>`. Maybe later we can search for variables with type `FilePath` or type `String` and names like `paths, files, dirs`. But GHC seems to have some problems with the `FilePath` type, so for now a more advanced approach won't be robust enough...

vrom911:
Resolves #275

### DIFF
--- a/src/Stan/Inspection/AntiPattern.hs
+++ b/src/Stan/Inspection/AntiPattern.hs
@@ -375,7 +375,7 @@ stan0214 = mkAntiPatternInspection (Id "STAN-0214") "use 'compare'" UseCompare
 -- | 'Inspection' — Slashes in paths @STAN-0215@.
 stan0215 :: Inspection
 stan0215 = mkAntiPatternInspection (Id "STAN-0215") "Slashes in paths" (FindAst pat)
-    & descriptionL .~ "Usage of '/' or '\' in paths results in the errors on different operation systems"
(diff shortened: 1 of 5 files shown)


src/Stan/Inspection/AntiPattern.hs
```diff
@@ -368,3 +371,20 @@ stan0214 = mkAntiPatternInspection (Id "STAN-0214") "use 'compare'" UseCompare
...
+        ]
+    & severityL .~ Error
+  where
+    pat :: PatternAst
+    pat =   opApp pathLit filepathOperator (?)
+        ||| opApp (?) filepathOperator pathLit
+
+    pathLit :: PatternAst
+    pathLit = PatternAstConstant (ContainStr "/")
+        ||| PatternAstConstant (ContainStr "\\")
```
chshersh, line 390: I think it should be 4 slashes... Just to have a more robust check.

vrom911: Good catch! 👍

PR #289, opened 2020-06-21
Fix cabal description

chshersh:


src/Stan/Inspection/AntiPattern.hs
```diff
@@ -329,7 +329,7 @@ filepathOperator = PatternAstName operatorPosix fun
     filePathType :: PatternType
     filePathType = "FilePath" `baseNameFrom` "GHC.IO" |:: []
         ||| stringPattern
-        -- ||| primTypeMeta "[]" |:: [ charPattern ]
```
chshersh, line 332: Haddock seems to have problems with this line. I've returned it back.

PR #396, opened 2020-10-09
🤖  Suggest `Text` and `ByteString` instead of `String`

### LINKED ISSUE #386: Suggest `Text` and `ByteString` instead of `String`
We discussed this with @vrom911 and decided that it would be nice to suggest more efficient string types in the following places:

1. Data type fields.
2. Function return arguments.

kutyel:
Fixes #386, happy Hacktoberfest! 🚀 

---

PS: sorry for the unrelated change of the `hie.yaml`, HSL didn't work in my local machine without this file

### DIFF
--- a/hie.yaml
+++ b/hie.yaml
@@ -1,0 +1,10 @@
+cradle:
+  stack:
+    - path: "./src"
+      component: "stan:lib"
+    - path: "./app"
+      component: "stan:exe:stan"
+    - path: "./target"
+      component: "stan:lib:target"
+    - path: "./test"
+      component: "stan:test:stan-test"
--- a/src/Stan/Analysis/Analyser.hs
+++ b/src/Stan/Analysis/Analyser.hs
@@ -25,7 +25,7 @@ import Stan.Hie (eqAst)
 import Stan.Hie.Compat (HieAST (..), HieFile (..), Identifier, NodeInfo (..), TypeIndex)
 import Stan.Hie.MatchAst (hieMatchPatternAst)
 import Stan.Inspection (Inspection (..), InspectionAnalysis (..))
-import Stan.NameMeta (NameMeta, ghcPrimNameFrom)
+import Stan.NameMeta (NameMeta, baseNameFrom, hieFindIdentifier, ghcPrimNameFrom)
 import Stan.Observation (Observations, mkObservation)
 import Stan.Pattern.Ast (Literal (..), PatternAst (..), anyNamesToPatternAst, case', constructor,
                          constructorNameIdentifier, dataDecl, fixity, fun, guardBranch, lambdaCase,
@@ -65,6 +65,7 @@ createVisitor hie exts inspections = Visitor $ \node ->
         BigTuples -> analyseBigTuples inspectionId hie node
         PatternMatchOn_ -> analysePatternMatch_ inspectionId hie node
         UseCompare -> analyseCompare inspectionId hie node
+        StringUsage -> analyseStringUsage inspectionId hie node
 
 {- | Check for big tuples (size >= 4) in the following places:
 
@@ -82,6 +83,23 @@ analyseBigTuples insId = matchAstWith isBigTuple insId tuple
     isBigTuple Node{..} = case nodeChildren of
         _:_:_:_:_  -> True
         _lessThan4 -> False
+
+{- | Check for usage of the 'String' type in the following places:
+
+* Type signatures: foo :: a -> String
+* Data type fields: newtype HasString = HasString { foo :: String }
+-}
+analyseStringUsage
+    :: Id Inspection
+    -> HieFile
+    -> HieAST TypeIndex
+    -> State VisitorState ()
+analyseStringUsage insId = matchAstWith usesString insId typeSig
+  where
+    usesString :: HieAST TypeIndex -> Bool
+    usesString = isJust . hieFindIdentifier stringName
+    stringName :: NameMeta
+    stringName = "String" `baseNameFrom` "GHC.Base"
 
 {- | Find usages of multiple comparison operators and suggest using
 'compare'. Currently, handles the following cases:
--- a/src/Stan/Analysis/Analyser.hs
+++ b/src/Stan/Analysis/Analyser.hs
@@ -30,7 +30,7 @@ import Stan.Observation (Observations, mkObservation)
 import Stan.Pattern.Ast (Literal (..), PatternAst (..), anyNamesToPatternAst, case', constructor,
                          constructorNameIdentifier, dataDecl, fixity, fun, guardBranch, lambdaCase,
                          lazyField, literalPat, opApp, patternMatchArrow, patternMatchBranch,
-                         patternMatch_, rhs, tuple, typeSig)
+                         patternMatch_, rhs, tuple, typeSig, type_)
 import Stan.Pattern.Edsl (PatternBool (..))
 
 import qualified Data.Map.Strict as Map
@@ -94,7 +94,7 @@ analyseStringUsage
     -> HieFile
     -> HieAST TypeIndex
     -> State VisitorState ()
-analyseStringUsage insId = matchAstWith usesString insId typeSig
+analyseStringUsage insId = matchAstWith usesString insId type_
   where
     usesString :: HieAST TypeIndex -> Bool
     usesString = isJust . hieFindIdentifier stringName
--- a/src/Stan/Analysis/Analyser.hs
+++ b/src/Stan/Analysis/Analyser.hs
@@ -25,7 +25,7 @@ import Stan.Hie (eqAst)
 import Stan.Hie.Compat (HieAST (..), HieFile (..), Identifier, NodeInfo (..), TypeIndex)
 import Stan.Hie.MatchAst (hieMatchPatternAst)
 import Stan.Inspection (Inspection (..), InspectionAnalysis (..))
-import Stan.NameMeta (NameMeta, baseNameFrom, hieFindIdentifier, ghcPrimNameFrom)
+import Stan.NameMeta (NameMeta, hieFindIdentifier, ghcPrimNameFrom, stringNameMeta)
 import Stan.Observation (Observations, mkObservation)
 import Stan.Pattern.Ast (Literal (..), PatternAst (..), anyNamesToPatternAst, case', constructor,
                          constructorNameIdentifier, dataDecl, fixity, fun, guardBranch, lambdaCase,
@@ -97,9 +97,7 @@ analyseStringUsage
 analyseStringUsage insId = matchAstWith usesString insId type_
   where
     usesString :: HieAST TypeIndex -> Bool
-    usesString = isJust . hieFindIdentifier stringName
-    stringName :: NameMeta
-    stringName = "String" `baseNameFrom` "GHC.Base"
+    usesString = isJust . hieFindIdentifier stringNameMeta
 
 {- | Find usages of multiple comparison operators and suggest using
 'compare'. Currently, handles the following cases:
--- a/src/Stan/Inspection/AntiPattern.hs
+++ b/src/Stan/Inspection/AntiPattern.hs
@@ -87,6 +87,7 @@ antiPatternInspectionsMap = fromList $ fmapToFst inspectionId
     , stan0213
     , stan0214
     , stan0215
+    , stan0216
     ]
 
 -- | Smart constructor to create anti-pattern 'Inspection'.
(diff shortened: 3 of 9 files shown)


target/Target/AntiPattern.hs
```diff
@@ -92,3 +92,9 @@ stanSlashesWindows' = "xxx" </> "asd\\asd"
 
 stanSlashesNo :: FilePath
 stanSlashesNo = "xxx" </> "asd"
+
+stanStringsAreBad :: String
+stanStringsAreBad = "please never use this data type"
+
+stanFunctionReturningString :: a -> String
+stanFunctionReturningString = const "nooo"
```
vrom911, line 100: Haha 😄 Like your tests! 💯

src/Stan/Inspection/AntiPattern.hs
```diff
@@ -388,3 +390,15 @@ stan0215 = mkAntiPatternInspection (Id "STAN-0215") "Slashes in paths" (FindAst
     pathLit :: PatternAst
     pathLit = PatternAstConstant (ContainStr "/")
         ||| PatternAstConstant (ContainStr "\\\\")
+
+-- | 'Inspection' - use 'Text' or 'ByteString'  instead of 'String' @STAN-0216@.
+stan0216 :: Inspection
+stan0216 = mkAntiPatternInspection (Id "STAN-0216") "String"
+  (FindAst $ PatternAstName strNameMeta (?))
```
vrom911, line 397: `PatternAstName strNameMeta (?)` is not exactly what is needed in here. This one would look for any `String` in the code.
In this case, we need to check that the `String` is in type, specifically, that it is the returned value. 

Maybe this example can be of use:
https://github.com/kowainik/stan/blob/43dd2c7cd27f7a5f729586d551438df9e5efa4f8/src/Stan/Inspection/Style.hs#L54-L67

In this example, we search for big tuples. For `String` we would need to do something similar as well. 

If that doesn't make sense, just ping us with any questions 🙂

hie.yaml
```diff
@@ -0,0 +1,10 @@
+cradle:
+  stack:
+    - path: "./src"
+      component: "stan:lib"
+    - path: "./app"
+      component: "stan:exe:stan"
+    - path: "./target"
+      component: "stan:lib:target"
+    - path: "./test"
+      component: "stan:test:stan-test"
```
vrom911, line 10: No worries! It could be handy for those who also use HLS 👌🏼

vrom911 (commented):
@kutyel Great start! 
I wrote a bit on the pattern itself, let me know if it makes sense, we could discuss it further 🙂

src/Stan/Inspection/AntiPattern.hs
kutyel: You mean to do something like this?
```haskell
stan0216 :: Inspection
stan0216 = Inspection
    { inspectionId = Id "STAN-0216"
    , inspectionName = "String"
    , inspectionDescription = "Usage of the default 'String' Haskell type"
    , inspectionSolution =
        [ "Consider using instead 'Text' for readable text, or 'ByteString' otherwise"
        ]
    , inspectionCategory = Category.antiPattern :| [Category.syntax]
    , inspectionSeverity = Style
    , inspectionAnalysis = StringUsage
    }
```
It's not really evident to me how to implement this `StringUsage` matcher, sorry 😖

chshersh: @kutyel Yeah, you need to introduce a separate constructor for the `InspectionAnalysis` type. `StringUsage` works fine :+1: 

You can see the implementation of `analyseBigTuples` to check how such custom check is implemented:

https://github.com/kowainik/stan/blob/43dd2c7cd27f7a5f729586d551438df9e5efa4f8/src/Stan/Analysis/Analyser.hs#L69-L84

kutyel: I arrived to the point in which I had to compare an `HieAST TypeIndex` to a "String", how can I do that? 🤔

kutyel:
Maybe the `hacktoberfest-accepted` label could be added to this PR? If it is not much to ask 😉

chshersh:
@kutyel Done!

src/Stan/Inspection/AntiPattern.hs
chshersh: @kutyel You need to implement a pattern to match the `String` type. I guess the most reliable way to figure out, how the pattern should look like, is to debug the content of the HIE files and see what's inside. For this you need to write a function you want to view and follow the instructions from our wiki on how to debug HIE files in GHCi:

* https://github.com/kowainik/stan/wiki/Development#view-hie-ast-of-a-file

src/Stan/Analysis/Analyser.hs
```diff
@@ -83,6 +84,23 @@ analyseBigTuples insId = matchAstWith isBigTuple insId tuple
...
+-}
+analyseStringUsage
+    :: Id Inspection
+    -> HieFile
+    -> HieAST TypeIndex
+    -> State VisitorState ()
+analyseStringUsage insId = matchAstWith usesString insId typeSig
+  where
+    usesString :: HieAST TypeIndex -> Bool
+    usesString = isJust . hieFindIdentifier stringName
```
kutyel, line 100: Apparently this function `hieFindIdentifier` is not doing what I expected 😕 .
There is no other usage in the code, maybe I'm using it wrong? I've confirmed that what I'm looking for are `nodeIdentifiers` that contain my given `NameMeta`...
A little more help on this would be greatly appreciated 🙏  @chshersh @vrom911

jfmengels:
It's great that you give a solution in the reported error. That said, I don't know Haskell, but as a beginner I think I would be confused as to why `String` is considered an anti-pattern when reading the error report. I would advise to add a short explanation and/or a link to a longer explanation. What do you think?

Whether to choose `Text` or `ByteString` seems less confusing though :+1:

src/Stan/Inspection/AntiPattern.hs
```diff
@@ -388,3 +391,17 @@ stan0215 = mkAntiPatternInspection (Id "STAN-0215") "Slashes in paths" (FindAst
...
         ||| PatternAstConstant (ContainStr "\\\\")
+
+-- | 'Inspection' - use 'Text' or 'ByteString'  instead of 'String' @STAN-0216@.
+stan0216 :: Inspection
+stan0216 = Inspection
+    { inspectionId = Id "STAN-0216"
+    , inspectionName = "String"
+    , inspectionDescription = "Usage of the default 'String' Haskell type"
+    , inspectionSolution =
+        [ "Consider using instead 'Text' for readable text, or 'ByteString' otherwise"
```
vrom911, line 402: I would split it into two separate suggestions 🙂 
To `ByteString` I would add instead of just otherwise, that it could be used for bytes or something like this. 
What you wrote is correct, and I would do the same at the beginning, but turns out people don't like when we talk about `ByteString` generally, so we need to be more careful in order not to trigger people and give them another reason "not to use Stan" 😄

src/Stan/Inspection/AntiPattern.hs
```diff
@@ -388,3 +391,17 @@ stan0215 = mkAntiPatternInspection (Id "STAN-0215") "Slashes in paths" (FindAst
     pathLit :: PatternAst
     pathLit = PatternAstConstant (ContainStr "/")
         ||| PatternAstConstant (ContainStr "\\\\")
+
+-- | 'Inspection' - use 'Text' or 'ByteString'  instead of 'String' @STAN-0216@.
+stan0216 :: Inspection
+stan0216 = Inspection
+    { inspectionId = Id "STAN-0216"
+    , inspectionName = "String"
```
vrom911, line 399: Let's provide a more descriptive name:
```suggestion
    , inspectionName = "Usage of the 'String' type"
```

src/Stan/Inspection/AntiPattern.hs
```diff
@@ -388,3 +391,17 @@ stan0215 = mkAntiPatternInspection (Id "STAN-0215") "Slashes in paths" (FindAst
...
+    , inspectionDescription = "Usage of the default 'String' Haskell type"
```
vrom911, line 400: And for the description, we can give a reason why we suggest this 🙂 (As @jfmengels suggests as well 👍🏼  )

target/Target/AntiPattern.hs
```diff
@@ -92,3 +92,11 @@ stanSlashesWindows' = "xxx" </> "asd\\asd"
...
+
```
vrom911, line 101: Let's also add another test, like

```haskell
stanFunctionTakingString :: String -> Bool
stanFunctionTakingString = const True
```

to make sure that it won't be triggered.

src/Stan/Analysis/Analyser.hs
```diff
@@ -83,6 +84,23 @@ analyseBigTuples insId = matchAstWith isBigTuple insId tuple
...
+    :: Id Inspection
+    -> HieFile
+    -> HieAST TypeIndex
+    -> State VisitorState ()
+analyseStringUsage insId = matchAstWith usesString insId type_
+  where
+    usesString :: HieAST TypeIndex -> Bool
+    usesString = isJust . hieFindIdentifier stringName
+    stringName :: NameMeta
+    stringName = "String" `baseNameFrom` "GHC.Base"
```
vrom911, line 102: We can add this to the `NameMeta` module and reuse in string pattern as well 🙂

vrom911 (commented):
Great work, @kutyel 🙏🏼 Thanks a lot for helping with this 👍🏼

src/Stan/Inspection/AntiPattern.hs
```diff
@@ -388,3 +391,17 @@ stan0215 = mkAntiPatternInspection (Id "STAN-0215") "Slashes in paths" (FindAst
...
+    , inspectionSolution =
+        [ "Consider using instead 'Text' for readable text, or 'ByteString' otherwise"
+        ]
+    , inspectionCategory = Category.antiPattern :| [Category.syntax]
+    , inspectionSeverity = Style
```
chshersh, line 405: This is actually a performance problem, not a stylistic one:

```suggestion
    , inspectionSeverity = Performance
```

src/Stan/Pattern/Type.hs
```diff
@@ -89,9 +89,7 @@ infix 5 |::
 -- | 'PatternType' for list @[a]@ or @'String'@.
 listPattern :: PatternType
 listPattern =
-    listNameMeta |:: [ (?) ]
-    |||
-    "String" `baseNameFrom` "GHC.Base" |:: []
+    listNameMeta |:: [ (?) ] ||| stringPattern
```
chshersh, line 92: I would keep patterns on separate lines for readability reasons :slightly_smiling_face: 
It makes it easier to see that the list pattern is a choice between two type patterns
```suggestion
    listNameMeta |:: [ (?) ]
    |||
    stringPattern
```

chshersh (commented):
Great work, @kutyel! It was nice v-meeting you and pair programming :hugs: 

I've also added a few comments in m review

kutyel:
Done, thanks for your reviews! ❤️

target/Target/AntiPattern.hs
kutyel: Why it should *not* be triggered if it is a function argument? 🤔

src/Stan/Analysis/Analyser.hs
```diff
@@ -83,6 +84,21 @@ analyseBigTuples insId = matchAstWith isBigTuple insId tuple
...
+-}
+analyseStringUsage
+    :: Id Inspection
+    -> HieFile
+    -> HieAST TypeIndex
+    -> State VisitorState ()
+analyseStringUsage insId = matchAstWith usesString insId type_
+  where
+    usesString :: HieAST TypeIndex -> Bool
+    usesString = isJust . hieFindIdentifier stringNameMeta
```
vrom911, line 100: I see that the test, when `String` is the first argument, is triggering the inspection.
I think this is because of this line. We need to check that it is either in uses String in the data type, or that the function returns it. It is going to be a bit more complicated check though in here. We can start with one check, for example with String in data types, and then deal with the returning String case 🙂

kutyel: Doesn't it make sense to warn agains all uses of `String`? I want to understand the logic behind only checkin the _last usage of String_ in the type signature. 😉

vrom911: Not always you can control the input of all functions, and lots of the functionality in the wild use `Strings`. But when we have custom functions, that we can control, it is recommended to move toward `Text` and return `Text`, that is the meaning for the last argument 🙂

PR #421, opened 2021-03-17
Replace deprecated set-env command

parry84:
The release workflow has been broken by removal of `set-env`, `add-path` commands from Github Actions: https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/

This changes replaced the removed commands with equivalent ones and upgrade dependency that where broken for the same reason.


tomjaguarpaw:
Thanks. All these changes have recently been incorporated.

PR #432, opened 2021-06-04
add Soisy to companies using stan

marcosh:


vrom911 (approved):
Thank you so much! ❤️

PR #506, opened 2023-09-01
Support up to and including GHC 9.4

tomjaguarpaw:
The approach is to make compatibility modules per individual GHC version.


tomjaguarpaw:
Hi @vrom911, is there any chance you'd be able to look at this? `stan` is a very useful tool and it would be great to have support for later GHCs available to the community. I am willing to put in the work and become a comaintainer if that would help you.

PR #521, opened 2023-10-14
Changes to avoid GHC warnings on building

mpilgrem:
These changes avoid almost all GHC warnings on building. The one warning I could not work out how to avoid was:
~~~text
<no location info>: warning: [-Wunused-packages]
    The following packages were specified via -package or -package-id flags,
    but were not needed for compilation:
      - base-4.17.2.0 (exposed by flag -package-id base-4.17.2.0)
~~~

### DIFF
--- a/src/Stan/Hie.hs
+++ b/src/Stan/Hie.hs
@@ -16,6 +16,7 @@ module Stan.Hie
     ) where
 
 import Colourista (errorMessage, infoMessage, warningMessage)
+import Prelude hiding (span)
 import System.Directory (doesDirectoryExist, doesFileExist)
 import System.Directory.Recursive (getDirRecursive)
 import System.FilePath (takeExtension)
--- a/src/Stan/Hie/MatchAst.hs
+++ b/src/Stan/Hie/MatchAst.hs
@@ -17,7 +17,7 @@ module Stan.Hie.MatchAst
     ) where
 
 import Data.Char (toLower)
-
+import Prelude hiding (span)
 import Stan.Core.List (checkWith)
 import Stan.Ghc.Compat (nameOccName, occNameString)
 import Stan.Hie (slice)
@@ -54,7 +54,7 @@ hieMatchPatternAst hie@HieFile{..} node@Node{..} = \case
            hieMatchPatternAst hie node p1
         && hieMatchPatternAst hie node p2
     PatternAstConstant lit ->
-           Set.member literalAnns (Set.map toNodeAnnotation (nodeAnnotations nodeInfo))
+           Set.member literalAnns (Set.map toNodeAnnotation (nodeAnnotations nodeInfo'))
         && ( let span = slice nodeSpan hie_hs_src in case lit of
                 ExactNum n   -> (span >>= readMaybe . decodeUtf8) == Just n
                 ExactStr s   -> span == Just s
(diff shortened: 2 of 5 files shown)


tomjaguarpaw:
Thanks! I'm not a big fan of making effort to avoid warnings, to be honest, but I'll take most of these. A couple I'd like to undo and rethink.

src/Stan/Hie/MatchAst.hs
```diff
@@ -54,37 +54,37 @@ hieMatchPatternAst hie@HieFile{..} node@Node{..} = \case
...
     PatternAstIdentifierDetailsDecl declType -> any (any (isDecl declType) . identInfo) $
-        Map.elems $ nodeIdentifiers nodeInfo
+        Map.elems $ nodeIdentifiers nodeInfo'
   where
     matchAnnotations :: Set NodeAnnotation -> NodeInfo TypeIndex -> Bool
     matchAnnotations tags NodeInfo{..} =
       tags `Set.isSubsetOf` Set.map toNodeAnnotation nodeAnnotations
 
-    nodeInfo = Stan.Hie.Compat.nodeInfo node
+    nodeInfo' = Stan.Hie.Compat.nodeInfo node
```
tomjaguarpaw, line 87: Would you mind undoing this one? Or alternatively add `import qualified Stan.Hie.Compat`.

src/Stan/Hie.hs
```diff
@@ -66,12 +66,12 @@ When the given source is empty returns 'Nothing'.
...
-slice span =
+slice span' =
     fmap
-        ( BS.take (srcSpanEndCol span - srcSpanStartCol span)
-        . BS.drop (srcSpanStartCol span - 1)
+        ( BS.take (srcSpanEndCol span' - srcSpanStartCol span')
+        . BS.drop (srcSpanStartCol span' - 1)
         )
-    . flip (!!?) (srcSpanStartLine span - 1)
+    . flip (!!?) (srcSpanStartLine span' - 1)
```
tomjaguarpaw, line 74: Please change this one into `import Preluding hiding (span)`

src/Stan/Pattern/Type.hs
```diff
@@ -115,36 +116,45 @@ integerPattern =
...
+#endif
 
+#if __GLASGOW_HASKELL__ >= 900
 naturalPattern900 :: PatternType
 naturalPattern900 = NameMeta
     { nameMetaName       = "Natural"
     , nameMetaModuleName = "GHC.Num.Natural"
     , nameMetaPackage    = "ghc-bignum"
     } |:: []
+#endif
```
tomjaguarpaw, line 157: Please undo the extra CPP here. I'd like to have a think whether there's a better way.

src/Stan/Hie/MatchAst.hs
mpilgrem: @tomjaguarpaw, that is not the source of the shadowing problem that the `'` fixes. The problem is that for `ghc-8.10.7` (and maybe other versions) `nodeInfo` is a field of the data constructor of type `HieTypes.HieAST a`.

src/Stan/Hie.hs
mpilgrem: OK.

src/Stan/Hie/MatchAst.hs
tomjaguarpaw: Ah I see. Let's keep this one then.

src/Stan/Pattern/Type.hs
mpilgrem: @tomjaguarpaw, you may also not like it, but so you can see what it looks like 'on the page', I've put the CPP condition around everything affected by it - which also allows for some simplification within each limb of the condition.

tomjaguarpaw: Yeah, I think that's the better way to go, but I'd rather get the slamdunks in first and then come back to the debatable cases later.

mpilgrem: OK.

mpilgrem:
@tomjaguarpaw, hopefully this now reflected what you were looking for.

tomjaguarpaw (approved):

tomjaguarpaw:
Thanks! Set to automerge. Feel free to make to submit the other changes as a follow-up PR. I'll give them some thought.

PR #522, opened 2023-10-14
Avoid GHC warnings about 'Defined but not used'

mpilgrem:
This solution uses a single `#if #elif #endif` (rather than the existing two) and moves all the affected code within each limb. It also allows for some simplication within each limb.


tomjaguarpaw (approved):

tomjaguarpaw:
OK, sure, this seems fine. Set to auto-merge.

PR #532, opened 2023-10-21
Support GHC 9.8

0rphee:
- Bump dependencies.
- Add Debug908.hs
- Fix renaming of `(,)` to `Tuple2` from `GHC.Tuple.Prim` (https://hackage.haskell.org/package/ghc-prim-0.11.0/docs/GHC-Tuple-Prim.html#t:Tuple2)
- Set conditional `--allow-newer` for ci


tomjaguarpaw:
Fantastic! We can release this as-is as version `0.1.0.2` and then when the dependencies are updated `stan` will automatically support GHC 9.8 with nothing extra required. (We can also remove the `--allow-newer` when the dependencies are updated.)

tomjaguarpaw:
Oops, I merged the wrong thing. Completing this as https://github.com/kowainik/stan/pull/533

PR #549, opened 2023-12-16
Bump stack.yaml to GHC 9.6.3/lts-22.0

mpilgrem:
Also preserves/updates `stack-ghc-9.4.8.yaml` for GHC 9.4.8, which may be useful for use of Stan in CI.


stan.cabal
```diff
@@ -17,6 +17,7 @@ build-type:          Simple
 stability:           experimental
 extra-doc-files:     README.md
                      CHANGELOG.md
+                     stack-ghc-9.4.8.yaml
                      stack.yaml
 extra-source-files:  test/.stan-example.toml
```
tomjaguarpaw, line 22: That's interesting. I wonder if this should be an `extra-source-file` rather than an `extra-doc-file` (and `stack.yaml` too). Thoughts?

mpilgrem: Files listed in `extra-source-files` are tracked by `cabal build` and changes in them cause (partial) rebuilds. Files in `extra-doc-files` are just additional files to be included in source distributions. I don't have a strong view, either way.

mpilgrem: `extra-doc-files` was introduced with Cabal 1.18 (September 2013). The discussion here https://github.com/commercialhaskell/stack/issues/2105 was in 2016 when, I assume, people were still supporting Cabal < 1.18.

mpilgrem: Actually, on balance, I think you are correct - it should be `exta-source-files`.

tomjaguarpaw (approved):

tomjaguarpaw:
Thanks!

PR #552, opened 2023-12-22
Break up runStan function and expose it and its components

0rphee:
This is no much more than anything a separation of functions, so that `stan` functionality can be better added to the `hls-stan-plugin` (https://github.com/haskell/haskell-language-server/pull/3914). 

`getStanConfig` is the only essential one for the plugin. I factored-out the second `getAnalysis` to use it as-is for `hls`. However, that is not possible because all the internal machinery in `stan` works with __only__ relative paths, while the plugin at the end requires an absolute path. I'm inclined to leave it as is, but if it were possible to change `stan` internals to get one version of the function with both the plugin and `stan` itself, that would be nicer.

This also exports `runStan`, but it would be no problem leaving it there.

### DIFF
--- a/src/Stan.hs
+++ b/src/Stan.hs
@@ -8,6 +8,9 @@ Main running module.
 
 module Stan
     ( run
+    , runStan
+    , getAnalysis
+    , getStanConfig
 
       -- ** Internal
     , createCabalExtensionsMap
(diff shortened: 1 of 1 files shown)


tomjaguarpaw (approved):

tomjaguarpaw:
Nice, thank you!

tomjaguarpaw:
Released to Hackage as https://hackage.haskell.org/package/stan-0.1.2.0

PR #557, opened 2024-02-05
Support base64-1.0

tomjaguarpaw:
I'd really rather not do this. I'd rather `extractBase64` be added to the `base64-0.4` so we can write code that works across both versions, but if we can't persuade the maintainer to do that we can fall back to this.

### DIFF
--- a/src/Stan/Observation.hs
+++ b/src/Stan/Observation.hs
@@ -1,3 +1,5 @@
+{-# LANGUAGE CPP #-}
+
 {- |
 Copyright: (c) 2020 Kowainik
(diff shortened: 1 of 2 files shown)


tomjaguarpaw:
Requested that `base64` add the compatibility layer itself in https://github.com/emilypi/base64/issues/59

tomjaguarpaw:
I'm going to include https://github.com/kowainik/stan/pull/557 in here, since we might as well support the bumped clay and base64 at the same time.

tomjaguarpaw:
Closing in favour of https://github.com/kowainik/stan/pull/559

PR #559, opened 2024-02-05
version -> 0.1.2.1

tomjaguarpaw:


tomjaguarpaw:
I'm going to include https://github.com/kowainik/stan/pull/557 in here, since we might as well support the bumped clay and base64 at the same time.

PR #566, opened 2024-05-31
Warning about overzealous -Wunused-packages

philderbeast:
I had a go a fixing this by only depending on base in components that needed it. After commenting out all other components except the main library, I found that base was needed and the `-Wunused-packages` warning was a false positive. Rather than removed the `-Wunused-packages` warning, I opted to leave a comment pointing to the relevant GHC issue.


tomjaguarpaw:
Thanks!

PR #569, opened 2024-05-31
Avoid -Wname-shadowing with Text.Blaze.Html5.s

philderbeast:
For the `ghc-9.10.1` build we can pick up `blaze-html-0.9.2.0` but with it get a `-Wname-shadowing` warning. This fixes that and I opted to rename `s` locally rather than use `CPP` with the import.

https://hackage.haskell.org/package/blaze-html-0.9.2.0/changelog

https://hackage.haskell.org/package/blaze-html-0.9.2.0/docs/Text-Blaze-Html4-FrameSet.html#v:s


tomjaguarpaw (approved):

tomjaguarpaw:
Thanks!

PR #581, opened 2025-01-17
Relax upper bound of hedgehog

alexfmpe:


tomjaguarpaw (approved):

tomjaguarpaw:
Revised as `-r1`: https://hackage.haskell.org/package/stan-0.1.3.0/revisions/

PR #585, opened 2025-03-13
Improve finding of Cabal files

Bodigrim:
Might fix https://github.com/haskell/haskell-language-server/issues/4515 or at least eliminate a few suspects.

### DIFF
--- a/src/Stan/Cabal.hs
+++ b/src/Stan/Cabal.hs
@@ -98,9 +98,8 @@ findCabalFileDir dir = do
     dirContent <- listDirectory dir
     let cabalFiles = filter isCabal dirContent
     pure $ case cabalFiles of
-        []          -> Nothing -- throwError $ NoCabalFile dirPath
-        [cabalFile] -> Just $ dir </> cabalFile
-        x:_xs       -> Just x -- throwError $ MultipleCabalFiles (x :| xs)
+        []            -> Nothing
+        cabalFile : _ -> Just $ dir </> cabalFile
   where
     isCabal :: FilePath -> Bool
     isCabal p = takeExtension p == ".cabal"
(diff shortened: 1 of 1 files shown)


Bodigrim:
@tomjaguarpaw any resolution on this?

tomjaguarpaw:
The Cabal file finding code looks very dubious and your suggestion improves it slightly. I have no objection to merging your suggestion per se, but since @TeofilC suggests that it doesn't help much I'd prefer to try to find the root cause before determining the correct resolution.

I'm happy to dig in myself to diagnose, if someone can give me a repeatable way of demonstrating this behaviour in HLS.

Bodigrim:
I think ideally Stan should not recursively traverse file system upfront at all. Instead, when checking a Haskell souce file in a particular folder, determine the corresponding Cabal file and cache the information about enabled extensions in `IORef (Map FilePath (Either ExtensionsError ParsedExtensions))`.

tomjaguarpaw:
As I understand it, if the path to the Cabal file is provided in `stanArgsCabalFilePath` https://github.com/kowainik/stan/blob/dc0a3a550dee53d51a018dda730d3d3631024dd7/src/Stan/Cli.hs#L57 then it *doesn't* traverse the filesystem. It only traverses the filesystem if that list is empty.

Bodigrim:
Probably `hls-stan-plugin` does not pass a list of Cabal files. Although I don't know if it's easy to determine them reliably on HLS side.

tomjaguarpaw:
I suspect you are right. See https://github.com/haskell/haskell-language-server/issues/4515#issuecomment-2780739873 for further analysis. I don't know exactly how HLS works, but regardless of who finds the cabal file, they should be doing it once, not multiple times.

tomjaguarpaw:
That's OK, I'm happy to shepherd it from here. My position is that I'll merge a fix to the root cause of the problem, if it can be determined. I'm not going to merge anything speculatively, in the mere hope that it will improve things.

Bodigrim:
Closing in favor of #586.

